PR #3, opened 2020-02-07
Varya Theme: Adding a css-variables version of the original version of Varia.

allancole:
image

This PR introduces a new version of Varia based on CSS-variables (aka custom properties). Combined with the [theme dev build-tool](https://github.com/Automattic/theme-dev-utils). The theme takes common theme design decisions and distills them down to a set of _design-tokens_ that can be altered by simply resetting a robust list of variables. It treats a theme design like a style guide which reduces the amount of hand-written CSS needed to produce a theme. 

It’s still a work in progress but currently it supports: 
- Starter Gutenberg styles for the frontend and backend.
- WooCommerce styles for the frontend and backend (plus WC blocks).
- Provides basic styling for non-Gutenberg area like Header, Footer, Comments, archives and legacy, pre-gutenberg styles.
- A custom theme accent color can be set via the Customizer and reflected in Gutenberg color palettes. (maybe this should be removed?) 

**To do:** 

- [ ] Integrate with the [child-theme build tool](https://github.com/Automattic/theme-dev-utils).
- [ ] Add Jetpack support?
- [ ] Optimize responsive font-size styles.
- [x] Audit how variables appear in the editor, the customizer, and the frontend.
- [ ] Consider adding defaults fallbacks to top-level css-variables.
- [x] Create a demo.
- [ ] Add a CSS-Variables polyfill or other fallback solution for older browsers.
- [ ] Create a ReadMe that describes how to use it.


jffng:
👏 

Haven't given this a full review yet, but one quick thing: 

When I tried incorporating the build tool, this line gave me an issue:

https://github.com/Automattic/themes-workspace/blob/6ac221dbefed63c7470ebd10d3af329b806774d2/variatheme/sass/abstracts/_imports.scss#L13

Removing the relative path fixed the issue (i.e. `sass/abstracts/responsive-logic` > `responsive-logic`). Was there a specific reason why this was written with the relative path?

allancole:
> Was there a specific reason why this was written with the relative path?

@jffng Looks like a simple copy/paste typo :-)

jffng:
As an FYI, to test this I'm [trying out](https://github.com/Automattic/themes-workspace/tree/try/montichello) building [Montichello](https://www.figma.com/file/Bs9FN7ryOPOcYafCy0weEm/Restaurants-and-Cafes%3A-Montichello?node-id=118%3A0&viewport=-618%2C36%2C0.29324716329574585).

I think it's working as expected:

- I can use the build tool to generate a child theme
- I can alter variables in the child theme to affect the output of the design

Since it pretty much works out of the box, I would highly encourage a readme to be the next step out of any of the to-dos. That will make iterating on this as a tool we use for theme development much more explicit. 

So far, most of the issues I'm encountering in development involve having to track down where and how the variables affect the design, and either change them (which could live in varia or the child theme) or write styles with a higher specificity in the child theme. Having a readme that outlines how the variables map to specific aspects of the design would help immensely.

allancole:
> I would highly encourage a readme to be the next step out of any of the to-dos. 

@jffng Yup, fully agreed here. Might actually be good to do daily ReadMe updates so its easier to keep track of changes/improvements. 

> I can use the build tool to generate a child theme

So this part of the tool is still incomplete actually. It’ll generate the child-theme but it isn’t connected properly which means its missing some of the necessary files and probably overwrites more CSS than it needs to. Yesterday, I committed a vary-child-theme example which you can copy and use as a better starting point than what the build tool can produce right now.

jeffikus:
I'm working through creating the child theme with build tool.

varya/functions.php
```diff
@@ -0,0 +1,306 @@
...
+		// Add support for full and wide align images.
+		add_theme_support( 'align-wide' );
+
+		// Add support for editor styles.
+		add_theme_support( 'editor-styles' );
+
+		// Do not enqueue editor styles here.
+		// Editor styles are loaded in `enqueue_block_editor_assets`
+		// See: varya_editor_theme_variables();
+		// add_editor_style( 'style-editor.css' );
```
kjellr: Is it weird to have this here at all if it's unused?

allancole:
> I'm also slightly confused about the `varya-child` and `varya-child-advanced` themes that are added in here... Are those just examples? In any case, I don't mind merging this and then adding documentation afterwards. This will be a work in progress, so having something merged in as a base is an important start.

Yup, I’m working out documentation in the ReadMe file as we speak. To put it simply though, `varya-child` is a basic child theme meant to help people get familiar with the variables and what changes when you override them. `varya-child-advanced` is an example of what an actual theme built with might Varya looks like. 

> Would it be possible to put those root-level Sass files somewhere else? Perhaps just directly in the sass folder? The root level of Varya is a little daunting right now because it's so full of files (and many of them share the same file name, but just different extensions).

Yup, this is one of the thing me and @jeffikus discusses in our last chat. The current build tool needs those files on the root level, but we can definitely change it. 

> A custom theme accent color can be set via the Customizer and reflected in Gutenberg color palettes. (maybe this should be removed?)

> We can probably strip this out later and migrate this functionality to global styles, right? If that's the case, I say we leave it in. If we envision technical problems with that, then we should strip it out for now.

Yup, thats actually already been stripped out [here](https://github.com/Automattic/themes-workspace/pull/3/commits/8dcae0afe03f5041349b65d64dc06f4b6e4a4e76) :-). Regarding, Global Styles, eventually we can do the same thing Go is starting to do where we rename our CSS-variables to the exact same names that will be in Global Styles.

varya/functions.php
allancole: Yup, theres quite a few areas that have notes and comments that can be removed. Figured we can do that kind of stuff in a final code review before submitting to wporg.

varya/README.md
```diff
@@ -0,0 +1,127 @@
+# Varya
+
+Varya is a styling system for quickly creating Gutenberg-ready themes for WordPress.
+
+## What does it do?
+
+When you reduce a theme design down to a set of systematic design decisions, you end up with something called a _Style Guide_. The Varya system works by taking the rules of a Style Guide and expressing them through carefully placed variables or _design tokens_ that influence the appearance of a WordPress site. It currently unifyies the overall of Gutenberg Blocks, the theme Header + Footer areas, WooCommerce, Jetpack and more. It also syncs styles between the editor and the frontend so that you don’t need to hand-write styles for both. This greatly speeds up the Gutenberg theme development process and reduces the amount of manual styling that typically goes into developing a theme.
```
kjellr: > It currently unifyies the overall of Gutenberg Blocks, the theme Header + Footer areas, WooCommerce, Jetpack and more.

This doesn't totally make sense. I think it's missing a word or two?

varya/README.md
```diff
@@ -0,0 +1,127 @@
+# Varya
+
+Varya is a styling system for quickly creating Gutenberg-ready themes for WordPress.
```
kjellr: Should this be rephrased to make clear that it's a parent theme itself?

varya/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+## What does it do?
+
+When you reduce a theme design down to a set of systematic design decisions, you end up with something called a _Style Guide_. The Varya system works by taking the rules of a Style Guide and expressing them through carefully placed variables or _design tokens_ that influence the appearance of a WordPress site. It currently unifyies the overall of Gutenberg Blocks, the theme Header + Footer areas, WooCommerce, Jetpack and more. It also syncs styles between the editor and the frontend so that you don’t need to hand-write styles for both. This greatly speeds up the Gutenberg theme development process and reduces the amount of manual styling that typically goes into developing a theme. 
+
+### What controls does the system come with?
+
+  - **Fonts** - Font-family, size, weight, and line-height rules. 
+  - **Colors** - Primary, secondary, background, foreground and border colors. 
+  - **Spacing** - Sets an 8px vertical rhythm between all blocks and major components. Adds general spacing rules for blocks and components. Adds utility spacing classes for negative margins.
```
kjellr:
```suggestion
  - **Spacing** - A default 8px vertical rhythm between all blocks and major components. It also includes utility spacing classes for negative margins.
```

varya/README.md
```diff
@@ -0,0 +1,127 @@
...
+  - **Responsive Logic** - Unifies the responsive behavior across various Blocks and Components to simplify behaviors.
```
kjellr:
```suggestion
  - **Responsive Logic** - Built-in responsive behavior across Blocks and Components.
```

varya/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Simple Child Theme Structure 
+(See: `/varya-child`)
+
+Use this simple Varya child-theme example to familiarize yourself with the CSS-variables and to see how they influence the theme design when you change them. To start, open up the `/varya-child-simple/variables.css` file. Change any of the values, save the file, and refresh the frontend of your site to see the changes. To add more variables, view the [full list of available variables](https://github.com/Automattic/themes-workspace/blob/add/variatheme/varya/variables.css) from the Varya parent them and copy any variables you want in to your child theme. 
+The stylesheet structure for a child-theme cascades downward like this:
+
+### Frontend
+- `/varya/variables.css`
+- `/varya-child-simple/variables.css` (System Override)
```
kjellr:
```suggestion
- `/varya-child-simple/variables.css` (System Overrides)
```

varya/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+A list of CSS-variables that override the variables in the child theme. This is where you tell the system to use the _Futura_ font-family instead of the _sans-serif_ default, for example. There’s no need to replace all of the variables here, only the ones you wish to actually change based on your theme design.
+
+**Varya parent theme font style** (`/varya/variables.css`)
+
+`:root { --global--font-family: sans-serif; }`
+
+**Varya child-theme font-style overrides** (`/varya-child/variables.css`)
+
+`:root { --global--font-family: Futura, Helvetica; }`
```
kjellr:
```suggestion
`:root { --global--font-family: "Futura", "Helvetica", sans-serif; }`
```

varya/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Extra CSS
+
+These should be supplemental styles that give the theme a unique appearance beyond what’s possible with the Varya system. Need to add a fixed header or add a box-shadow to your theme’s buttons? This is where those styles would go. When possible and appropriate, try to include Varya CSS-variables in this styles sheet so that the system retains its usefulness across the theme.
```
kjellr:
```suggestion
These should be supplemental styles that give the theme a unique appearance beyond what’s included with the Varya system defaults. Need to add a fixed header or add a box-shadow to your theme’s buttons? This is where those styles would go. When possible and appropriate, try to use Varya CSS-variables in this stylesheet so that the system retains its usefulness across the theme.
```

varya/README.md
```diff
@@ -0,0 +1,127 @@
...
+ - Optimize responsive font-size styles.
+ - [DONE] Audit how variables appear in the editor, the customizer, and the frontend.
+ - Consider adding defaults fallbacks to top-level css-variables.
+ - Create a demo.
+ - Add a CSS-Variables polyfill or other fallback solution for older browsers.
+ - [WIP] Create a ReadMe that describes how to use it.
+
+### License
+
+- GPL 2.0
```
kjellr: This should link to the LICENSE file. Also, are there any other license disclosures that we should list here?

varya/README.md
```diff
@@ -0,0 +1,127 @@
...
+**Varya child-theme font-style overrides** (`/varya-child/variables.css`)
+
+`:root { --global--font-family: Futura, Helvetica; }`
+
+### Extra CSS
+
+These should be supplemental styles that give the theme a unique appearance beyond what’s possible with the Varya system. Need to add a fixed header or add a box-shadow to your theme’s buttons? This is where those styles would go. When possible and appropriate, try to include Varya CSS-variables in this styles sheet so that the system retains its usefulness across the theme.
+
+## Advanced Child Theme Structure 
+(See: `/varya-child-advanced`)
```
jffng: I feel like this should eventually be a separate example with its own repo?

varya/README.md
```diff
@@ -0,0 +1,127 @@
...
+**Frontend** 
+- `/varya/variables.css`
+- `/varya/style.css`
+
+**Editor** 
+- `/varya/variables-editor.css`
+- `/varya/style-editor.css`
+
+**Customizer**
+- `/varya/variables.css`
```
jffng: Is there a reason the customizer variables and frontend variables are contained in the same partial?

varya/README.md
```diff
@@ -0,0 +1,127 @@
...
+  - Search for: `'varya-child'` and replace with: `'my-theme-name'` (with quotes).
+  - Search for: `varya_child_` and replace with: `my_theme_name_`.
+  - Search for: `Text Domain: varia-child` and replace with: `Text Domain: my-theme-name` in _style.css_.
+  - Search for:  `varia-child` and replace with: `my_theme_name`.
+  - Search for: `varia-child-` and replace with: `my-theme-name-`.
+
+Soon this process will be replaced by a `theme-dev-util` (similar to [this](https://github.com/Automattic/theme-dev-utils)) that allows you to run a command and automatically produce a child theme with all the strings already replaced.
+
+## Simple Child Theme Structure 
+(See: `/varya-child`)
```
jffng: I may not understand fully, but isn't the child theme the actual theme you activate on your site? If so, should that be made more explicit?

varya/README.md
kjellr: Yeah, that's a good idea. But it's fine to keep this in here for now.

varya/README.md
allancole: Yup, will do!

varya/README.md
kjellr: Actually... I think these are all missing license files. 😄

allancole: Ahh, will do. 

May need to reference the image used in the screenshot here. Need to double check that it’s CC0, first but that would likely go here. I’ll compare with a few themes to see what else might need to be included here.

varya/README.md
```diff
@@ -0,0 +1,127 @@
...
+- `/varya/variables-editor.css`
+- `/varya/style-editor.css`
+
+**Customizer**
+- `/varya/variables.css`
+- `/varya/style.css`
+
+In each view, the variables are loaded first and then the stylesheet is loaded which applies the variables.
+
+## How to use it
```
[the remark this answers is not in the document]
allancole: Fully agreed. 

I need to sync with @jeffikus to get the exact process as it relates to the build tool, but even with out the tool, the steps here should be more explicit than they are now. Good catch and will do.

varya/README.md
allancole: Yeah, this is actually just the Varya version of “Nwaneri” so eventually it’ll be its own thing anyway.

Alternatively, I wonder if common code snippets might be useful to include here. With Varya themes you spend most of your time working on the header & footer UI since those areas are Gutenberg-able yet. 

If we were to come up with 2-3 common header patterns for example, we could create some Varya specific snippets that give you a jump start on styling those areas, at least until FSE is here.

varya/README.md
allancole: Woops! Yeah, that was left over from when Varya still had customizer color options. Now that those are gone, I can remove this section.

varya/README.md
allancole: Yup, once I revise the “How To Use It” section this should become more clear.

jffng:
Merging so we can iterate on Botany / a default style.

PR #7, opened 2020-03-10
Add a barebones base for block-based themes 

kjellr:
A very barebones starting point for building block-based themes. Powered by CSS variables. 

Screen Shot 2020-03-10 at 3 16 42 PM

---

Still todo: 

- [x] Get variables working in the editor
- [ ] Strip out unused CSS
- [x] Strip out unused PHP


kjellr:
Should be ready to mess around with. I expect we'll add and adjust other variables over time — this is just a relatively empty base. 

One annoying caveat I've noticed so far: The editor styles do not work with `experimental-theme.json` out of the box, so I've included an additional `variables.css` file that duplicates those rules. :/ This should be just temporary though.

nielslange:
> Spaces to tabs! 🙃

https://www.youtube.com/watch?v=SsoOG6ZeyUI 😁

kjellr:
This is in a generally workable state, so I'm just going to go ahead and merge it in so that we can easily test it out by trying out child themes. We'll continue iterating from there. 😅

blockbase/block-template-parts/footer.html
```diff
@@ -0,0 +1,3 @@
+<!-- wp:paragraph {"align":"center","fontSize":"small"} -->
+<p class="has-text-align-center has-small-font-size">Proudly powered by <a href="https://wordpress.org">WordPress</a></p>
```
nielslange: Just out of curiosity, what's the benefit of naming the classes `has-text-align-center` and `has-small-font-size` over  `text-align-center` and `font-size-small`?

kjellr: That all comes straight from Gutenberg. 🙂

nielslange: Ah, I see. Thanks for clarifying this, @kjellr. 🙌

PR #8, opened 2020-03-10
Montichello using minimal block-based theme and global styles support

jffng:
**Description**
This is a fork of #7, expanding that example to build a version of Montichello that supports both the existing block-based themes and global styles implementation. 

Screen Shot 2020-03-10 at 5 41 52 PM

**To-do**
- [x] Convert Varya CSS variables to the supported global styles CSS variables
- [ ] Update `experimental-theme.json` and `css/variables.css` with correct values for the theme design
- [ ] Update the `block-templates` and `block-template-parts` according to the theme design


jffng:
Currently the only difference from @kjellr's PR is that this PR:

- loads an extra set of block styles for the front end
- registers block style variations using the server side method

kjellr:
I added a few small updates. Still a bit messy, but here's the latest: 

Screen Shot 2020-03-11 at 12 48 42 PM

kjellr:
I'm going to close this one for now since we're moving this repo over to [automattic/themes](https://github.com/automattic/themes). I'll leave the branch intact in case we need to take a look at it in the future.

PR #9, opened 2020-03-18
[Blockbase] Cleanup + CSS restructuring

kjellr:
This PR adds a readme with basic documentation for Blockbase. Also a little copyediting. 

It also does a little reorganization to the CSS files: 

- The CSS files are split up into their own `front-end` and `editor` folders for clarity.
- There are new `style-custom.css` and `editor-custom.css` files for people to use for theme-specific overrides. This should allow most of the theme CSS to stay untouched, allowing for most structural CSS files to be replaced over time if blockbase were to be updated.
- Cleaned up CSS file headers with descriptions of what each file does.


blockbase/css/front-end/style-base.css
```diff
@@ -0,0 +1,436 @@
...
+iframe,
+object {
+	max-width: 100%;
+}
+
+/*--------------------------------------------------------------
+# Custom Styles
+--------------------------------------------------------------*/
+
+@import url( "./css/custom.css" );
```
jffng: I think this import is no longer needed given how you've restructured.

blockbase/css/front-end/style-blocks.css
```diff
@@ -1,3 +1,10 @@
+/*--------------------------------------------------------------
+# Base Block Styles
+#
+# These provide a baseline for front-end block styles. 
+# Please override these in custom.css.
```
jffng:
```suggestion
# Please override these in style-custom.css.
```

blockbase/css/front-end/style-base.css
```diff
@@ -0,0 +1,436 @@
+/*--------------------------------------------------------------
+# Base Styles
+#
+# These provide a baseline for all front-end styles. 
+# Please override these in custom.css.
```
jffng:
```suggestion
# Please override these in style-custom.css.
```

blockbase/css/front-end/style-base.css
kjellr: Good catch! Should be all set now.

PR #10, opened 2020-03-18
[Blockbase] Remove duplicate call to enqueue editor styles

jffng:
**Description**

This call to the enqueue the editor styles within the theme setup causes an error and is also unnecessary. 

This PR removes it.


kjellr:
Ah, right. The weird thing is that this removed line is actually the _right_ way to call editor styles. 

The enqueue function later on is not the way we're really supposed to do it. That second call is necessary only to the styles working in the full-site editor. I think a better way of doing this might be to keep this call in place, but limit the second call for styles to _only_ the full-site editor. There are examples for this in the comments here: 

https://developer.wordpress.org/reference/hooks/admin_enqueue_scripts/

kjellr:
> However when I load style-editor.css using add_editor_style(), I get a 404 due to the imports:

Interesting! Turns out `@import` will always do that if you enqueue via `add_editor_style()`. To get that working, I removed the imports and instead inserted an array in `add_editor_style()`. Try that out and let me know if it works for you.

jffng:
That fixed it, thanks!

PR #14, opened 2020-03-23
Adding Varya

jffng:
**Description**

Let's add `varya` and `varya-child` theme to the main repository so we can iterate on them.


jffng:
Removing `varya-child` as to keep things as lean as possible for now.

PR #15, opened 2020-03-24
Implement CSS Variables fallback support

jffng:
**Description**

Currently this PR adds a postcss plugin that finds all custom properties and turns them into static values, placing a "duplicate" style rule before the custom property.  

**Testing Instructions**

- Check out this PR
- `npm i`
- `npm run build`
- Verify that the styles have not been affected

To-do:
- [x] Add to the package build scripts
- [ ] Test against IE11
- [ ] Explore conditional loading for specific browsers


varya/style.css
```diff
@@ -364,16 +400,30 @@ body[class*="woocommerce"] #page .woocommerce button.button,
...
 	cursor: pointer;
+	font-weight: normal;
 	font-weight: var(--button--font-weight);
+	font-family: 'Fira Sans'
+,
+Helvetica
+,
+Arial
+,
+sans-serif;
```
allancole, line 416: @jffng, something weird is happening here and I’m not sure why but it seems like an issue with `postcss-custom-properties`. This likely happens in a few other places where strings or CSS values have commas. Might be worth it to run the post-css linting _after_ the `postcss-custom-properties` process so it gets fixed automatically, but I’m not sure that’s possible.

varya/style.css
```diff
@@ -39,52 +38,51 @@ Nicolas Gallagher and Jonathan Neal https://necolas.github.io/normalize.css/
...
+		--responsive--aligndefault-width: calc(482px - var(--responsive--spacing-horizontal));
+		--responsive--alignwide-width: calc(482px - var(--responsive--spacing-horizontal));
 		--responsive--alignright-margin: calc( 0.5 * (100vw - var(--responsive--aligndefault-width)));
 		--responsive--alignleft-margin: calc( 0.5 * (100vw - var(--responsive--aligndefault-width)));
 	}
 }
 
-@media only screen and (min-width: 640px) {
+@media only screen and (min-width: 592px) {
 	:root {
```
allancole, line 56: Looks like the responsive logic won’t work at all in ie11 even with this plugin :-/ 

The `--responsive--align****-width` variables are updated on `:root` based on the viewport size. This greatly reduces the amount of CSS needed to achieve responsiveness, but it prevents the fallbacks from appearing where they need to be for the PostCSS plugin. 

To get this to work, we may need to refactor and introduce an `%extend` to group all the responsive rules together so the plugin can process the resulting values properly. This comes with the downside of adding a fairly big chunk of new CSS responsive with their accompanying fallbacks. We could also refactor this with a normal `@mixin`, but that would produce even more new CSS lines in an already pretty large CSS file. 

Cc: @kjellr as this might be a blocker to providing full ie11 support if the suggested solutions don’t work.

kjellr: Ok. I think we should de-prioritize this part of the issue for now. We can come back to this closer to launch. 👍

allancole: Sounds good.

kjellr:
I'm going to close this one for now since we're moving this repo over to [automattic/themes](https://github.com/automattic/themes). I'll leave the branch in place in case we want to pull from it later on.

PR #18, opened 2020-03-27
[Varya] Switch to using add_editor_style to enqueue block editor stylesheets.

kjellr:
We should be using [the standard way](https://developer.wordpress.org/block-editor/developers/themes/theme-support/#editor-styles) of enqueing the block editor stylesheets: `add_editor_style`. 

I'm not sure why Varya was using `enqueue_block_editor_assets` originally, but I'm not noticing any obvious issues swapping between this PR and the original one — the editor looks identical. 

Assuming there's no other technical limitation that requires use of `enqueue_block_editor_assets`, this just needs a little testing to ensure it's bringing in all the appropriate stylesheets still.

### DIFF
--- a/varya/assets/css/style-editor.css
+++ b/varya/assets/css/style-editor.css
@@ -269,7 +269,7 @@
 	margin-top: var(--global--spacing-vertical);
 }
 
-#wpwrap .editor-styles-wrapper {
+body {
 	color: var(--global--color-foreground);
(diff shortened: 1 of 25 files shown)


jffng:
Going to merge this so I can rebase the default styles integration against it, thanks!

kjellr:
Thanks, folks!

PR #19, opened 2020-03-27
Try adding a block-based theme version of Varya

kjellr:
Just a quick experiment to see if we could concurrently build a block-based version of Varya alongside the one we're building in here today. 

This version exists as a child theme to keep things really simple. It inherits all the front-end and editor styles it's supposed to, and adds in just a handful of block templates and styling to get it working. Looks pretty good, with very little effort: 

**Front End:**
Screen Shot 2020-03-26 at 10 34 50 PM

**Full Site Editor:** 
Screen Shot 2020-03-26 at 10 46 46 PM

---

Interested in your thoughts, @allancole + @jffng.


varya/functions.php
```diff
@@ -308,6 +308,6 @@ function varya_editor_theme_variables() {
 /**
  * Load default styles for Varya parent-theme-only
  */
-if ( ! is_child_theme() ) {
+//if ( ! is_child_theme() ) {
 	require get_template_directory() . '/default-style/functions.php';
```
jffng: This will be a separate PR, but did we decide to move `default-styles` into Varya itself?

Asking again, because this wouldn't needed if we either a) integrated the `default styles` directly into Varya or b) moved them into this child theme.

kjellr: Yes, I believe we did decide to move the default styles into Varya itself. The "vanilla" styles can either live as an alternate set of files/rules in there for now too (essentially the opposite of the current approach), or we can spin them out into a separate child theme. Let's open another issue to track that.

jffng: https://github.com/Automattic/themes-workspace/issues/20

kjellr:
This should be ready to go as an initial commit. Mind giving it one more look?

jffng:
>The main unfinished part is the fact that there's a little more vertical spacing than necessary between items in the header

kjellr:
> I removed the spacer blocks for now to tighten up the header.

Ah sorry — the spacer blocks were fine, it's more the automatic vertical margins in between the image, site title, and site description. But that's ok, we'll work on cleaning this all up later anyway.

Issue #20, opened 2020-03-27
[Varya] Make Botany styles the primary styles


jffng:
Currently, Varya conditionally loads a set of default styles if it is the active theme (i.e. no child theme is active):

https://github.com/Automattic/themes-workspace/blob/ebfa32c07c6f8261bd0949ee887cd07802e0eb34/varya/functions.php#L311-L313

We'd like to move these defaults to the base theme itself, so that the the theme structure is more conventional.

I believe this involves a few things:

- [ ] Archiving the vanilla styles and / or swapping them with the default styles
- [x] Changing CSS variables to match Botany
- [x] Modifying block specific sass to match Botany

This should happen before we make additional changes to the theme styles.

jffng:
Closing this as #21 takes care of most of it. 

We can address the vanilla styles in a subsequent PR when we revisit the child theme build tool.

PR #21, opened 2020-03-27
Integrate default styles into Varya

jffng:
To partially address #20, this PR makes it so the default styles we had started are the base theme styles. 

Test this by checking out the PR and verifying that the in-progress Botany styles are working as previously. 

I also snuck in a change that made the heading font-weights normal.


varya/assets/css/variables-editor.css
```diff
@@ -16,43 +17,43 @@
...
-	--global--font-size-md: calc(var(--global--font-size-base) * var(--global--font-size-ratio));
-	--global--font-size-lg: calc(var(--global--font-size-base) * var(--global--font-size-ratio) * var(--global--font-size-ratio));
-	--global--font-size-xl: calc(var(--global--font-size-base) * var(--global--font-size-ratio) * var(--global--font-size-ratio) * var(--global--font-size-ratio));
-	--global--font-size-xxl: calc(var(--global--font-size-base) * var(--global--font-size-ratio) * var(--global--font-size-ratio) * var(--global--font-size-ratio) * var(--global--font-size-ratio));
-	--global--font-size-xxxl: calc(var(--global--font-size-base) * var(--global--font-size-ratio) * var(--global--font-size-ratio) * var(--global--font-size-ratio) * var(--global--font-size-ratio) * var(--global--font-size-ratio));
-	--global--font-size-xxxxl: calc(var(--global--font-size-base) * var(--global--font-size-ratio) * var(--global--font-size-ratio) * var(--global--font-size-ratio) * var(--global--font-size-ratio) * var(--global--font-size-ratio) * var(--global--font-size-ratio));
+	--global--font-size-xs: 14px;
+	--global--font-size-sm: 18px;
+	--global--font-size-md: 20px;
```
allancole: Super close here, but the `font-size-md` variable needs to align with the base font size (18px). Otherwise, the entry-meta and the body copy will be the same font-size, when it should be smaller: 

image

If we make `font-size-sm` 15px, we can push all the other size up one level, so the max is 64px. We’ll probably have to drop the `xxxxl` 72px rule and only use that for the Block or Heading that actually uses it.

jffng:
@allancole I adjusted the font size variables per https://github.com/Automattic/themes-workspace/pull/21#discussion_r399566092 and dropped `xxxxl`. 

Is there anything else that needs addressed before merging this?

varya/assets/sass/components/header/_header-branding.scss
```diff
@@ -42,3 +42,21 @@
 	font-family: var(--branding--description--font-family);
 	font-size: var(--branding--description--font-size);
 }
+
+a {
+	text-decoration: none;
+	border-bottom: 1px solid var(--global--color-secondary);
+}
```
allancole: @jffng This is a tricky one. At first, I was thinking of moving it to `_reset.scss` where there’s already an `a {}` selector, but the bigger issue is that if we leave this in the Varya system, then every child theme that _doesn’t_ need an underline on anchor links will have to overwrite this (even when using the “Vanilla” variables). 

Wondering if there’s a way to have some styles like this one exist outside of the system in a way that allows child themes to not have to overwrite. 

This might also be minor enough that it’s not worth over-thinking in which case, I’d suggest just moving it to `_reset.scss` and undoing it in the Vanilla variables starter stylesheet. 

Cc: @kjellr

jffng: Per our convo in slack, I think moving it to `_reset.scss` makes sense in this case.

PR #22, opened 2020-03-27
Varya: Fix line-heights in editor and optimize editor stylesheet.

allancole:
- Optimizes the editor styles to remove a bunch of structural, header and footer rules that aren’t relevant for the editor. 
- Fixes the editor selectors so that the correct line-height for `--global--line-height-body` value is applied.
- Fixes a small typo that broke responsive styles on larger screens.

Before:
image

After: 
image

### DIFF
--- a/varya/assets/css/style-editor.css
+++ b/varya/assets/css/style-editor.css
@@ -1,281 +1,14 @@
 /**
  * These styles should be loaded by the Block Editor only
  */
-/**
- * Layout
- * - Structral and responsive styles
- */
-/**
- * Repsonsive Styles
- */
-/**
- * Required Variables
- */
-/**
- * Root Media Query Variables
- */
-:root {
-	--responsive--spacing-horizontal: calc(2 * var(--global--spacing-horizontal));
-	--responsive--aligndefault-width: 100%;
-	--responsive--alignwide-width: 100%;
-	--responsive--alignfull-width: 100%;
-	--responsive--aligndefault-width: 100%;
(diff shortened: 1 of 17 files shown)


varya/assets/css/style-editor.css
```diff
@@ -1,285 +1,20 @@
...
 	font-family: var(--global--font-secondary);
 	font-size: var(--global--font-size-root);
 	font-weight: normal;
-	line-height: var(--global--font-line-height-body);
 	-moz-osx-font-smoothing: grayscale;
 	-webkit-font-smoothing: antialiased;
 }
 
+p {
+	line-height: var(--global--line-height-body);
```
kjellr: I'm not sure this is actually necessary anymore – I tried turning this off in the inspector and things looked the same as they did with it enabled.

kjellr: Also a sidenote: this line height value should be `1.7`, not `1.78`. Would you mind updating that in this PR since you're working on line height anyway?

allancole: > I'm not sure this is actually necessary anymore – I tried turning this off in the inspector and things looked the same as they did with it enabled.

Yes but without it, if a child-theme has a different line-height, than they won’t be able to change it. 

> This line height value should be 1.7, not 1.78. 

Yup, will do. Afterward check to see that you you notice a change when enabling/disabling `line-height: var(--global--line-height-body);` like you did above—you should see a subtle change there :-)

PR #25, opened 2020-03-30
[Varya] Delete unused legacy sass files.

kjellr:
These were previously not imported anywhere, so they are not needed.


kjellr:
> I think this ‘legacy’ directory can be removed too

Updated in 35e3a6d. Mind retesting?

allancole:
Yup, looks good. 

LGTM!

PR #26, opened 2020-03-30
[Varya] Update site header styles

kjellr:
(This is based on #21, so that should get merged in first)

Adds a few quick styles to get the site header looking more like the comps: 

- Center-aligns the items in there.
- Adjust the vertical margins.
- Adds a circular mask to the site logo.
- Changes the font weight for the header to Bold. 
- Switches the underline approach to use borders. We'll want to make this a little cleaner, but this is decent for now. 

---

Before:

Screen Shot 2020-03-30 at 12 45 29 PM

After:

Screen Shot 2020-03-30 at 12 54 04 PM

### DIFF
--- a/varya/assets/sass/components/header/_config.scss
+++ b/varya/assets/sass/components/header/_config.scss
@@ -4,7 +4,7 @@
 	--branding--color-link-hover: var(--global--color-primary-hover);
 	--branding--title--font-family: var(--global--font-primary);
 	--branding--title--font-size: var(--heading--font-size-h1);
-	--branding--title--font-weight: normal;
+	--branding--title--font-weight: 700;
 	--branding--description--font-family: var(--global--font-secondary);
 	--branding--description--font-size: var(--global--font-size-sm);
 	--branding--description--font-family: var(--global--font-secondary);
--- a/varya/assets/sass/components/header/_config.scss
+++ b/varya/assets/sass/components/header/_config.scss
@@ -9,8 +9,8 @@
 	--branding--description--font-size: var(--global--font-size-sm);
 	--branding--description--font-family: var(--global--font-secondary);
 
-	--branding--logo--border-radius: 50%;
 	--branding--logo--max-width: 128px;
+	--branding--logo--max-height: 128px;
 
 	--primary-nav--font-family: var(--global--font-secondary);
 	--primary-nav--font-size: 16px;
--- a/varya/assets/sass/components/header/_config.scss
+++ b/varya/assets/sass/components/header/_config.scss
@@ -9,6 +9,9 @@
 	--branding--description--font-size: var(--global--font-size-sm);
 	--branding--description--font-family: var(--global--font-secondary);
 
+	--branding--logo--border-radius: 50%;
+	--branding--logo--max-width: 128px;
+
 	--primary-nav--font-family: var(--global--font-secondary);
 	--primary-nav--font-size: 16px;
 	--primary-nav--font-weight: normal;
--- a/varya/assets/css/style-woocommerce-rtl.css
+++ b/varya/assets/css/style-woocommerce-rtl.css
@@ -129,6 +129,11 @@ body[class*="woocommerce"] .entry-content > .woocommerce {
 	}
 }
 
+.site-header {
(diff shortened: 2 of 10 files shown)


varya/assets/sass/components/header/_config.scss
```diff
@@ -4,11 +4,14 @@
...
 	--branding--title--font-family: var(--global--font-primary);
 	--branding--title--font-size: var(--heading--font-size-h1);
-	--branding--title--font-weight: normal;
+	--branding--title--font-weight: 700;
 	--branding--description--font-family: var(--global--font-secondary);
 	--branding--description--font-size: var(--global--font-size-sm);
 	--branding--description--font-family: var(--global--font-secondary);
 
+	--branding--logo--border-radius: 50%;
+	--branding--logo--max-width: 128px;
```
jffng: I think these variable additions make sense.

kjellr:
> Do we want to tackle the primary navigation styles & position in a separate PR?

Yeah, I thought I'd handle these separately, since they might get a little more complicated. 👍

varya/assets/sass/components/header/_config.scss
allancole: Just noting that when we use the `--branding--logo--max-width: 128px;` variable, it needs to match the what appears in functions.php here: https://github.com/Automattic/themes-workspace/blob/update/site-branding/varya/functions.php#L88-L89. And its probably worth adding `--branding--logo--max-height` as well. 

I’m a little unsure about this variable `--branding--logo--border-radius: 50%;` and having the site-logo cropped by default. This probably deserves a deeper discussion/solution, but users have tons of different kinds of logos and if someone uploads a type-driven logo, they won’t understand why it gets cropped into a circle and also won’t have a way to directly edit it. To keep that from happening it may be better to remove the CSS cropping and create a pre-cropped PNG for the logo instead. That way users can pretty much upload anything and it’ll just display it as-is. 

We’ll have to deal with this differently with FSE, but for the customizer option, I think we’ll be better off leaning toward more flexibility.

kjellr: > I’m a little unsure about this variable --branding--logo--border-radius: 50%; and having the site-logo cropped by default. This probably deserves a deeper discussion/solution, but users have tons of different kinds of logos and if someone uploads a type-driven logo, they won’t understand why it gets cropped into a circle and also won’t have a way to directly edit it. To keep that from happening it may be better to remove the CSS cropping and create a pre-cropped PNG for the logo for the theme instead. That way users can pretty much upload anything and it’ll just display it as-is.

I'm not _totally_ sure about this one. It's a design choice by the theme, and I think it's okay to leave it in. I've removed it for now just to get this through quickly though.

kjellr:
Should be ready for another set of 👀!

jffng:
LGTM. Going to merge since @allancole's comments about the border radius and max-height were addressed.

varya/assets/sass/components/header/_config.scss
allancole: Wait, I think I misspoke in the previous comment :-). I like the circular logo as a design choice—it should definitely stay circular for the demos and theme screenshots in this theme. 

I’m only suggesting we implement `site-logo` in a different way that offers users more visual variety so they aren’t _forced_ to have circular logo if they don’t want one. Users don’t always understand the dynamics of the logo graphics they have. They’ll take their logo as-is, upload it to the customizer and when ([this happens](https://user-images.githubusercontent.com/709581/77959986-854fef00-72a5-11ea-9936-fa950fc8956d.png)) they won’t understand why or how to fix it—it just looks broken to them. If we remove the circular crop, _any_ image can be a logo, which is what users expect (even though it often comes out looking bad). 

Another thing to consider is that `--branding--logo--border-radius: 50%` suggests that a child theme might want to change the 50% to something else like 4px, or 33%. But when it comes to displaying the logo, I can only imagine 50% or 0 being useful values in the context of Varya. If only two values would ever be used there, and those values aren’t tied to nay other, similar design decisions, then it doesn’t necessarily _need_ to be a part of the system. It can just be normal CSS and be just as effective in a child theme or the parent theme.

Issue #24, opened 2020-03-27
Varya: Integrate spacing support in Layout Grid block


allancole:
So I’ve been testing the Layout Grid block against Varya, and it seems to work pretty well out of the box. I don’t think we really need to add extra support, however the spacing rules in the plugin are different than the ones in Varya. For instance: 

Layout Grid:
```
.wp-block-jetpack-layout-grid.wp-block-jetpack-layout-gutter__medium {
    grid-gap: 16px;
}
```

If we were to do integrate that style into Varya, it would look like this:
```
.wp-block-jetpack-layout-grid.wp-block-jetpack-layout-gutter__medium {
    grid-gap: var(--global--spacing-unit);
}
```

Could be overkill, but it would allow the spacing system in Varya to influence the paddings and gutters in the Layout Block. I’m anticipating us using this block pretty regularly so having the spacing values synced here means they’ll align with other spacing rules that are already in use in the system.

If we do this we’d also need to add a few more utility classes and CSS-variables to support the spacing options that come with the plugin. 

```
--global--spacing-none: 0px;
--global--spacing-small: ??px;
--global--spacing-medium: 16px; // same as --global--spacing-unit
--global--spacing-large: ??px;
--global--spacing-huge: ??px;
```

Curious, if you all think this is worth the effort (which wouldn’t be a lot of effort) @jffng @kjellr

jffng:
The layout grid seems worth supporting to me. Would this just mean adding another directory in `assets/sass/block/layout-grid` along with a couple sass files?

allancole:
> Would this just mean adding another directory in assets/sass/block/layout-grid along with a couple sass files?

Precisely. To do so, you’d just need to go through [the block’s CSS](https://plugins.trac.wordpress.org/browser/layout-grid/trunk/blocks/front.css), and pick out the padding/margin/gutter rules replace them with the appropriate Varya CSS-variables—but only when it makes sense. Feel free to take a stab at in a PR :-)

jffng:
Solved by #29.

PR #29, opened 2020-03-30
[Varya] Add Basic Support for the Layout Grid block

jffng:
**Description**
This PR addresses #24: adding support for the [Layout Grid block](https://wordpress.org/plugins/layout-grid/), tying it to the spacing system of Varya. As a first pass, it just ties everything to the `--global--spacing-unit` variable. 

Screen Shot 2020-03-30 at 4 53 16 PM

**To Test**
- Install the layout grid plugin
- Create a new post and add a layout block
- Add some content
- Make it full screen, try the various gutter sizes

**Known issues**
- _Front End_: The code uses `!important` to override the styles injected by the plugin. Could use a better solution here. 
- _Editor_: Nesting a layout grid inside of a group (e.g. as many of the dotcom starter page templates do) causes overflow issues 
- _Editor_:There is a visual discrepancy between the content and the grid overlay when the gutter size is set to Huge and the "Add end gutters" option is active.


varya/assets/sass/blocks/layout-grid/_style.scss
```diff
@@ -0,0 +1,17 @@
+@import 'config';
+
+.wp-block-jetpack-layout-grid {
+    grid-gap: var(--layout-grid--gutter-large) !important;
+
+    @each $size in $layout-gutter-sizes {
+        &.wp-block-jetpack-layout-gutter__#{$size} {
+            grid-gap: var(--layout-grid--gutter-#{size}) !important;
```
allancole: Yup, as you pointed out, let’s work out another way to solve this without `!important` as the last step before merging.

varya/assets/sass/blocks/layout-grid/_editor.scss
```diff
@@ -0,0 +1,30 @@
+@import 'config';
```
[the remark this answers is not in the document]
allancole: On second thought, this actually might be perfectly fine! As long as the mixin itself isn’t called in this file, only the `$layout-gutter-sizes` variable will get imported which is all we need. 👍

varya/assets/sass/blocks/layout-grid/_style.scss
```diff
@@ -0,0 +1,17 @@
+@import 'config';
```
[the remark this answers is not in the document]
allancole: Ignore this, as mentioned here: https://github.com/Automattic/themes-workspace/pull/29#discussion_r400521493

allancole:
I think this is working pretty well now. @jffng or @kjellr, can either of you all review this and test it against some more complex layout-grid content? 

If everything checks out, I’ll remove the `!important` in favor of something less aggresive as the last step before merging.

allancole:
Also noting here that I noticed the Layout-Grid plugin inserts a separate stylesheet into the footer of the HTML doc which is why those `!important` flags are needed in the first place. The ideal fix, would be to have the plugin enqueue those styles into the header where they usually go. 

I opened an issue to suggest that change here: https://github.com/Automattic/block-experiments/issues/64

varya/assets/sass/blocks/layout-grid/_style.scss
```diff
@@ -1,29 +1,40 @@
 @import 'config';
 
 .wp-block-jetpack-layout-grid {
-	grid-gap: var(--layout-grid--gutter-default);
+	grid-gap: var(--layout-grid--gutter-large) !important;
+	padding-left: var(--layout-grid--gutter-large) !important;
```
jffng, line 5: The addition of this padding causes the content to be misaligned in the editor, from my testing: 

**With padding:**
Screen Shot 2020-04-01 at 10 57 19 AM

**Without padding:** 
Screen Shot 2020-04-01 at 10 58 54 AM

Not sure if removing this padding causes other issues.

allancole: > The addition of this padding causes the content to be misaligned in the editor, from my testing:

Ahh, good catch! meant to separate that out, but I’ll do that now and look into the other issue you found as well. 

Thanks!

varya/assets/sass/blocks/layout-grid/_editor.scss
```diff
@@ -1,30 +1,106 @@
...
+			padding-left: calc(var(--layout-grid--gutter-#{$size}) + var(--layout-grid--background-offset));
+		}
+	}
+}
+
+/* Overlay grid */
+.wp-block-jetpack-layout-grid {
+	/* wpcom-overlay-grid is the classname targeting the grid overlay visual aid displayed in the editor */
+	.wpcom-overlay-grid {
+		grid-gap: var(--layout-grid--gutter-large);
```
jffng, line 76: Just noting that the way the plugin works now is that it defaults to the large gutter size. If this ever changes, our styles would break.

jffng:
This is working well for me, thanks for those changes! 

When you nest a layout grid inside a group block, there are overflow issues. This is the case for example on the dotcom starter page layouts. I think this might be outside the scope of what we should address in the theme styles, though, and rather open up an issue in the layout grid plugin itself. 

Screen Shot 2020-04-01 at 6 16 43 PM

@allancole what do you think?

varya/assets/sass/blocks/layout-grid/_editor.scss
allancole: Yeah this (and many other things) will likely change quite a bit as it grows and matures and Varya will have to grow with it. The good part about it is that we’d only have to fix it here once and every child theme we build on the system will inherit the fix mostly automatically :-) 

I also thing the Sass can be audited/cleaned up even further so its easier to adapt to changes too.

allancole:
Yeah, I think we should deal with the nested group block stuff separately. 

I’m also just realizing this Layout Block plugin is a JetPack thing. 

We may want to move the `assets/sass/blocks/layout-block/` directory to `assets/sass/vendors/layout-block/` since its not a core block.

jffng:
@allancole sounds good to me. 

Any ideas around how to make these selectors more specific given that the stylesheet for this block is injected at the bottom?

allancole:
@jffng Made another improvement in 128806b so that each vendor gets its on exclusive directory. Let me know what you think :-)

jffng:
@allancole Nice, makes sense so we can isolate issues as they arise. 🚢

Issue #27, opened 2020-03-30
Varya: Consider removing duplicate post meta on single post page. 


kjellr:
We currently have two lines of Meta content for each post on single post pages. When there isn't much content (or when there's no post title) this looks weird. @allancole: are there any potential downsides to us just removing the top one? 

It looks like we have separate (identical?) functions for each of those: [`varya_entry_meta()`](https://github.com/Automattic/themes-workspace/blob/daf833b07ace29008e0b2b8cfae0382cd694a628/varya/inc/template-tags.php#L70-L133) and [`varya_entry_footer()`](https://github.com/Automattic/themes-workspace/blob/daf833b07ace29008e0b2b8cfae0382cd694a628/varya/inc/template-tags.php#L135-L198). Maybe we can just delete the first one? 

---

Current: 

![Screen Shot 2020-03-30 at 1 02 23 PM](https://user-images.githubusercontent.com/1202812/77940493-c258b900-7286-11ea-9a56-e62d4d86ffe8.png)

Recommended: 

![Screen Shot 2020-03-30 at 1 02 09 PM](https://user-images.githubusercontent.com/1202812/77940508-c97fc700-7286-11ea-9d31-06b34dfbe3cd.png)

allancole:
> are there any potential downsides to us just removing the top one?

No downsides really. They are both there intentionally, but how its implemented is definitely worth an audit. Since there isn’t a Block for post meta (yet), I wanted there to be an easy way for child themes to do whatever they want here. Some child themes may want the meta above the content and others below, others may want author credits above, and only comment links below. 

Right now, child themes can achieve that be adding their own `varya_entry_meta()` or `varya_entry_footer()` functions in `themes/child-theme-name/functions.php` which will allow them to replace whats in the parent with any function in a non-destructive way. 

If we take out just the top one (without a function or an action to hook in to), then it means child themes will need to override all the `content-???.php` templates in the parent entirely, creating more complexity in the child theme than I think its worth. 

It’d probably be better to empty the top function and replace it with an action so child themes can easily put stuff there if they want to, but still have it empty by default for Varya.

kjellr:
Fixed in #30

PR #30, opened 2020-03-31
[Varya] Update post meta

jffng:
This PR addresses removes underlines on meta links and addresses #27 

**Before**:
Screen Shot 2020-03-31 at 11 37 21 AM

**After**:
Screen Shot 2020-03-31 at 11 37 03 AM

### DIFF
--- a/varya/style-rtl.css
+++ b/varya/style-rtl.css
@@ -2824,8 +2824,8 @@ nav a {
 }
 
 .site-logo .custom-logo {
-	border-radius: var(--branding--logo--border-radius);
 	max-width: var(--branding--logo--max-width);
+	max-height: var(--branding--logo--max-width);
 	height: auto;
 }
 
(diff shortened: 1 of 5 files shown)


varya/style-rtl.css
```diff
@@ -2824,8 +2824,8 @@ nav a {
 }
 
 .site-logo .custom-logo {
-	border-radius: var(--branding--logo--border-radius);
 	max-width: var(--branding--logo--max-width);
+	max-height: var(--branding--logo--max-width);
```
kjellr: Is this left over from some other PR? (Also, I think that should be `var(--branding--logo--max-height)`)

jffng: Ugh yes I think there was an instance where I failed to check in the compiled rtl styles, and this is leftover from that.

jffng: Ah actually this is a bug in https://github.com/Automattic/themes-workspace/blob/2f0d972bca1914f38b12bc7dc3aec5883ef676f9/varya/assets/sass/components/header/_header-branding.scss#L65

PR #31, opened 2020-03-31
Varya: Fix logo height variable + recompile CSS

kjellr:
Fixes the logo height variable as noticed [here](https://github.com/Automattic/themes-workspace/pull/30#pullrequestreview-384862581).

### DIFF
--- a/varya/style-rtl.css
+++ b/varya/style-rtl.css
@@ -2825,7 +2825,7 @@ nav a {
 
 .site-logo .custom-logo {
 	max-width: var(--branding--logo--max-width);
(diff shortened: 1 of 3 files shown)


varya/style-rtl.css
```diff
@@ -3271,6 +3271,7 @@ nav a {
 
 .entry-meta a,
 .entry-footer a {
+	border-bottom: none;
```
kjellr: This appears to be leftover from a previous commit. But I think it's supposed to be in here.

jffng: Yep, noticed this as well. Thanks for checking it in.

PR #32, opened 2020-03-31
[Varya] Refine quote and pullquote block styles

jffng:
This PR adjusts the quote and pullquote block styles to match the design comps. 

**Before**
Screen Shot 2020-03-31 at 2 05 02 PM

**After**
Screen Shot 2020-03-31 at 2 04 42 PM


varya/assets/css/variables-editor.css
```diff
@@ -125,16 +125,18 @@ body {
 	--list--font-family: var(--global--font-secondary);
 	--definition-term--font-family: var(--global--font-primary);
 	--pullquote--font-family: var(--global--font-primary);
-	--pullquote--font-size: var(--heading--font-size-h4);
-	--pullquote--border-width: 4px;
-	--pullquote--color-border: var(--global--color-foreground);
+	--pullquote--font-size: var(--heading--font-size-h2);
+	--pullquote--font-style: italic;
+	--pullquote--line-height: var(--global--line-height-heading);
```
jffng: I added two variables here, not entirely sure if they're necessary.

varya/assets/css/variables.css
```diff
@@ -35,7 +35,7 @@
 	/* Line Height */
 	--global--line-height-base: 1;
 	--global--line-height-body: 1.7;
-	--global--line-height-heading: 1.125;
+	--global--line-height-heading: 1.3;
```
jffng: I adjusted the value of this global variable to match the H1 - H3 line height from the comps.

kjellr:
Any chance we can adjust the top and bottom margins here so that they appear about the same as the left/right margins on the block?

Screen Shot 2020-03-31 at 3 11 54 PM

I think that's because of the line height of the quote, but we should be able to compensate for it somehow.

jffng:
I made the following changes:
- Removed the border and padding on right-aligned quote blocks:
Screen Shot 2020-04-01 at 4 12 13 PM

- Refined the alignwide and alignfull treatments on the pullquote block:
Screen Shot 2020-04-01 at 4 13 43 PM

- Matched the editor styles and front-end styles for those variations

varya/assets/css/variables.css
allancole: Coool. Makes sense 👍

varya/assets/css/variables-editor.css
allancole: Hmmm, I like the `line-height` variable. The italic one might be useful too, but just need to be careful that it doesn’t interfere with any inline-italic rules set by the users (which may be a problem already).

varya/assets/css/style-editor.css
```diff
@@ -819,37 +821,51 @@ p.has-background:not(.has-background-background-color) a {
...
 }
 
 .wp-block-quote.is-large p, .wp-block-quote.is-style-large p {
-	font-family: var(--quote--font-family);
-	font-size: var(--heading--font-size-h3);
-	letter-spacing: var(--heading--letter-spacing-h3);
-	line-height: var(--global--line-height-h3);
+	font-size: var(--quote--font-size-large);
+	font-style: normal;
+	line-height: 1.6;
```
allancole, line 853: Does line-height need to be connected to a variable here?

jffng: It should be. 

I connected it to `--global--line-height-body`, which is 1.7 even though the comps suggest 1.6 for this specific element. 

We can revisit this after #37.

jffng:
> Right-aligned block quotes lose the border treatment:

@kjellr I may have misinterpreted this feedback. Are the right-aligned block quotes supposed to retain a right border?

kjellr:
> I may have misinterpreted this feedback. Are the right-aligned block quotes supposed to retain a right border?

Yep! Sorry that wasn't clear. They should retain the single-pixel red border, just have it swapped over to the right side.

jffng:
Cool, I restored the right aligned border. This is ready for another review.

kjellr:
Looks good! I noticed a very minor bug: font sizes for the citations should be `--global--font-size-xs` instead of `--global--font-size-sm`. I've made that change, and will merge this in. 👍

PR #34, opened 2020-03-31
[Varya] Rework max width for content blocks in the editor

jffng:
Gutenberg injects a style rule that limits the max-width of blocks that are not align full or align wide to 580px: https://github.com/WordPress/gutenberg/blob/34253056f822cf554437d5ad449b033d811e199f/packages/edit-post/src/style.scss#L104

This causes inconsistency between the front-end and editor in this theme.

This PR leverages the responsive rules developed for the front end into the editor to set the default content width of blocks that are not align full or align wide. 

**Before**
Screen Shot 2020-03-31 at 5 54 06 PM

**After**
Screen Shot 2020-03-31 at 5 53 43 PM

If this approach is acceptable, then we should probably revisit how we implement the align wide responsive styles in the editor.

### DIFF
--- a/varya/assets/sass/base/_editor.scss
+++ b/varya/assets/sass/base/_editor.scss
@@ -1,3 +1,5 @@
+@import '../structure/responsive-logic';
+
 body {
 	color: var(--global--color-foreground);
 	background-color: var(--global--color-background);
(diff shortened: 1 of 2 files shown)


varya/assets/sass/base/_editor.scss
```diff
@@ -22,11 +24,23 @@ a {
...
 button,
 a {
 	cursor: pointer;
 }
+
+// Gutenberg injects a rule that limits the max width of .wp-block to 580px
+// This line overrides it to use the responsive spacing rules for default width content 
+.wp-block {
+	&:not([data-align="full"]):not([data-align="wide"]){
+		@extend %responsive-aligndefault-width;
```
allancole, line 40: The left/right margins are causing nested blocks conflicts. 

If you replace the `%extend` with just the width setting, it’ll produce the right widths without the conflicting margin rules:
```
max-width: var(--responsive--aligndefault-width);
```

jffng: Ahh gotcha. Do you think this is a reasonable place to put these rules for now?

PR #40, opened 2020-04-01
Varya: Update dropdown arrows to match comps

kjellr:
We need to update the dropdown arrows to match the comps. This first pass uses a quick CSS-only approach, but needs to be tested to see how it might scale if the font size/padding of the menu were to change in child themes. 

Another option would be to use something like the existing [`varya_add_dropdown_icons()`](https://github.com/Automattic/themes-workspace/blob/master/varya/inc/template-functions.php#L206-L250) function to inject a SVG in there instead (If we do not use that function, we should probably delete it). 

---

Before: 

Screen Shot 2020-04-01 at 3 43 20 PM
After:

Screen Shot 2020-04-01 at 3 42 28 PM

### DIFF
--- a/varya/assets/sass/components/header/_primary-navigation.scss
+++ b/varya/assets/sass/components/header/_primary-navigation.scss
@@ -196,11 +196,17 @@
 		& > div > ul > .menu-item-has-children > a {
 
 			&::after {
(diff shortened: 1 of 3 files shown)


kjellr:
Closing in favor of #42

Issue #41, opened 2020-04-01
Varya: Perform an audit of bundled SVG icons


kjellr:
We'll want to do an audit of [the SVG icons](https://github.com/Automattic/themes-workspace/blob/baaa2864700df921269eaf0c23a590c85c48b6c1/varya/classes/class-varya-svg-icons.php#L73-L175) bundled with the theme, to make sure they're all: 

- Used
- GPL compatible
- Noted in our license file when applicable

kjellr:
Adding a note here: We should ensure that the social icons reuse the same SVGs in Gutenberg's Social Links block, which should theoretically be GPL licensed (We can [look at Twenty Twenty](https://github.com/WordPress/twentytwenty/blob/6d0a5240af108d02f58ec797fd49b32458a4c698/classes/class-twentytwenty-svg-icons.php#L173-L256) to find those + mirror [the credit in that theme's README](https://github.com/WordPress/twentytwenty/blob/master/README.md)).

kjellr:
This is all set now that #52 and #65 are merged. 👍

PR #42, opened 2020-04-01
Varya: Update dropdown arrows to match comps (alternative)

allancole:
Props to @kjellr for getting this PR started. It’s similar to whats proposed in https://github.com/Automattic/themes-workspace/pull/40 but it uses SVG instead of the rotated corner border trick. 

- Adds SVG dropdown icons _only_ to top-level menu items.
- It also uses `color: currentColor` to simplify the way menu link colors are handled on :hover.

Before|After
------------|------------
image|

On the After image, notice how the hover color persists for the parent menu items. As the mouse moves to deeper nested menu-links, each parent menu-item retains its hovered state color. This isn’t a commonly used technique, but I like that it gives a subtle visual indication of the relationships between the nested links at different levels. Once this is synced with `—` m-dashes being removed in #39 , I think this subtlety might be nice to have.


kjellr:
> On the After image, notice how the hover color persists for the parent menu items. As the mouse moves to deeper nested menu-links, each parent menu-item retains its hovered state color. This isn’t a commonly used technique, but I like that it gives a subtle visual indication of the relationships between the nested links at different levels. 

I think this is actually a little confusing, since it is abnormal hover behavior. Since all sub-items are laid out visually anyway, that nested hover color isn't necessary to convey that child items are children. 

If we can eliminate that (I wonder if the CSS just needs more specificity?), I think this should be ready to go.

kjellr:
I adjusted the styles to get this working as I described above: 

hovers

Some of the specificity is a little weird, but it works. @jffng would you mind giving this another set of eyes?

kjellr:
> Visually, the arrow weight is a little heavy

Definitely. I'll tackle this as part of the work on #41. 

Thanks all!

PR #43, opened 2020-04-02
[Varya] Refine site title underline

jffng:
This PR adjusts the site title underline to avoid descenders, addressing this [issue](https://github.com/Automattic/themes-workspace/projects/1#card-35496673).  It uses `background-image` to draw the underline, based on [this reference](https://css-tricks.com/styling-underlines-web/). 

**Before:**
Screen Shot 2020-04-02 at 4 16 46 PM

**After**
Screen Shot 2020-04-02 at 2 59 08 PM

The tricky part is getting the text-shadows in balance with the "line-height"

### DIFF
--- a/varya/assets/sass/components/header/_header-branding.scss
+++ b/varya/assets/sass/components/header/_header-branding.scss
@@ -50,7 +50,22 @@
 		&:hover {
 			color: var(--branding--color-link-hover);
 		}
+
+		&::selection {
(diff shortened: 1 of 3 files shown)


kjellr:
This looks great! The descender from the g is just slightly messy, but I think we should just tackle that separately — this is such a big improvement over what we have currently. 👍

PR #44, opened 2020-04-02
Varya: Add new line-height control support

allancole:
While this works, and addresses both the frontend and backend pretty well. I’m not sure it’s the best approach. 

Demo: https://cloudup.com/cyNcpgnlaSp

**ToDo**
- [x] Audit line-height variable overrides to make sure it works consistently for gutenberg and doesn’t negatively impact non-gutenberg variables.
- [x] Consider how child-themes might be impacted by this change.
- [ ] Look into setting a default line-height for headings and paragraphs. The editor shows a `1.5` line-height by default despite what the CSS actually has as the default before changing the option.

### DIFF
--- a/varya/assets/css/style-editor.css
+++ b/varya/assets/css/style-editor.css
@@ -590,6 +590,15 @@ object {
 	line-height: var(--global--line-height-body);
 }
 
+.wp-block-heading h1[style*="--wp--typography--line-height"], h1[style*="--wp--typography--line-height"], .h1[style*="--wp--typography--line-height"],
+.wp-block-heading h2[style*="--wp--typography--line-height"], h2[style*="--wp--typography--line-height"], .h2[style*="--wp--typography--line-height"],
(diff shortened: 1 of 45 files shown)


allancole:
> For now, I think trying to scope the changes to the individual block level is best. For example, could we do something like:

> ```
> p {
>    line-height: var(--wp--typography--line-height, --global--line-height-body);
> }
> ```

Yeah, this feels like a better approach. I’ll give it a shot 👍

allancole:
@jffng I rebased this against `master` and tried your suggestion above—works like a charm. LGTM, if you agree :-)

allancole:
> Look into setting a default line-height for headings and paragraphs. The editor shows a 1.5 line-height by default despite what the CSS actually has as the default before changing the option.

I don’t think there’s a solution for this yet in Gutenberg. but I think we can merge this without it.

Issue #49, opened 2020-04-03
[Varya] Refine Post Comments


jffng:
Currently the post comments section looks like this: 
<img width="500" alt="Screen Shot 2020-04-03 at 12 19 20 PM" src="https://user-images.githubusercontent.com/5375500/78382823-6e5f1480-75a5-11ea-8bf8-db1cd7836dbf.png">

It should look like this:
<img width="500" alt="Screen Shot 2020-04-03 at 12 16 16 PM" src="https://user-images.githubusercontent.com/5375500/78382719-466fb100-75a5-11ea-8bad-d68c956968e5.png">

kjellr:
This should be all set now that #53 was merged.

PR #53, opened 2020-04-07
Varya: Refine the post comments

jffng:
This PR updates the existing post comments section to be more in line with design comps (#49). 

_Before_ 

Screen Shot 2020-04-07 at 4 20 11 PM

_After_

Screen Shot 2020-04-07 at 4 19 15 PM
Screen Shot 2020-04-07 at 4 18 58 PM

### DIFF
--- a/varya/assets/css/variables-editor.css
+++ b/varya/assets/css/variables-editor.css
@@ -81,8 +81,8 @@ body {
 	--form--line-height: var(--global-line-height-body);
 	--form--color-text: var(--global--color-foreground);
 	--form--border-color: var(--global--color-border);
-	--form--border-width: 1px;
-	--form--border-radius: 3px;
+	--form--border-width: 2px;
+	--form--border-radius: 0;
 	--form--spacing-unit: calc(0.33 * var(--global--spacing-unit));
 	/* Blocks */
 	--button--color-text: var(--global--color-background);
--- a/varya/assets/css/variables-editor.css
+++ b/varya/assets/css/variables-editor.css
@@ -83,7 +83,7 @@ body {
 	--form--border-color: var(--global--color-border);
 	--form--border-width: 2px;
 	--form--border-radius: 0;
-	--form--spacing-unit: calc(0.33 * var(--global--spacing-unit));
+	--form--spacing-unit: calc(0.5 * var(--global--spacing-unit));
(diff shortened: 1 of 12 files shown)


allancole:
Couple more small things I noticed :-)

Let’s make the form line-hieght match the body line-height. This looks too tight: 
image

“Cancel Reply” seems like its floating here. Maybe it should be centered or bottom-aligned with the “Leave a Comment” text? cc: @kjellr for design direction
image

jffng:
@allancole @kjellr Thanks for checking this out! I took another pass and made the following changes: 

- Logged out state
- Alignment of the author information
- Line height and padding of the form fields
- Slight adjustment of the "Cancel Reply" state

Screen Shot 2020-04-08 at 3 26 41 PM

I also added a bottom border, to avoid long lists of comments bleeding together:

Screen Shot 2020-04-08 at 3 27 55 PM

allancole:
I say LGTM! 👍

kjellr:
Made a small update to the borders:

- Removed the topmost border, so there's no border between the "Comments" header and the first comment.
- Used a lighter color for the border, so it's not as bold. 

Screen Shot 2020-04-09 at 2 51 21 PM

@allancole I used `--global--color-border-dark` here, and I also lightened that color in `_config.scss`. What are all these `light`/`dark` variations used for? The color I picked is also used in `--global--color-background-light`, and I'm not sure if that's a problem.

allancole:
> I used --global--color-border-dark here, and I also lightened that color in _config.scss. What are all these light/dark variations used for? The color I picked is also used in --global--color-background-light, and I'm not sure if that's a problem.

Yeah, I just took another look at those border variables and they’re not as intuitive as they probably should be. 

In the same way that `--global--color-foreground` has a `-light` and `-dark` variation, I figured border colors could benefit from variations too, but so far it’s been kind of a rare thing to use. They’ve been in the system for a while, but I don’t think I’ve ever actually changed them. 

I think it’s worth removing the variations and just keeping the one border-color. I’ll make an issue for it 👍

kjellr:
> I think it’s worth removing the variations and just keeping the one border-color. I’ll make an issue for it

Cool, sounds good. In the meantime, I'll merge this in. 👍

Issue #55, opened 2020-04-08
Varya: Left-aligned Cover Block margin issue


jffng:
This is what a left-aligned cover block looks with some content in line:

<img width="1121" alt="Screen Shot 2020-04-08 at 11 05 16 AM" src="https://user-images.githubusercontent.com/5375500/78800706-be651f00-7989-11ea-97f2-8be067c93ea3.png">

There is a responsive CSS rule on the cover block setting `margin-right: var(--responsive--alignleft-margin)`, and de-activating that rule appears to solve the issue:

<img width="1176" alt="Screen Shot 2020-04-08 at 11 14 50 AM" src="https://user-images.githubusercontent.com/5375500/78801073-2f0c3b80-798a-11ea-8faa-77e41f4d749e.png">

However, I'm not entirely sure what the intention behind this particular responsive variable is, so want to double check with @allancole.

PR #58, opened 2020-04-09
Varya: Fix left/right-aligned Cover Block margin issue.

allancole:
Fixes an issue where left/right aligned cover blocks will spill out of the text column.

Before|After
-----------|-----------
image|image

 The alignment class for Cover sits directly on the element like this: `.entry-content > .wp-block-cover.alignright/left` element. Other blocks like button or image nest the alignment class like this: `.entry-content > .wp-block-button > .alignright/left`. This means the utility class has to applied differently for each condition. This fix overrides the utility class so the Cover block is treated slightly different but still works as expected as far as the system goes.

- Fixes: #55

### DIFF
--- a/varya/assets/sass/blocks/cover/_style.scss
+++ b/varya/assets/sass/blocks/cover/_style.scss
@@ -75,6 +75,26 @@
 	/**
 	 * Block Options
 	 */
+	.entry-content > &.alignright {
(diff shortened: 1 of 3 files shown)


allancole:
Yup, this will likely change as Gutenberg evolves :-)

Issue #57, opened 2020-04-09
Varya: Remove global border-color variants


allancole:
It’s becoming apparent that the two `-light` and `-dark` color variations for border-color are rarely used, if at all. We should consider removing them for clarity. 

https://github.com/Automattic/themes-workspace/blob/1d69ab9eadc0c8e809c7f88263b506d70fdf31f7/varya/assets/sass/abstracts/_config.scss#L46-L48

jffng:
Noting that the following two variables are duplicates at the same level of hierarchy (`global`). 

https://github.com/Automattic/themes-workspace/blob/1d69ab9eadc0c8e809c7f88263b506d70fdf31f7/varya/assets/sass/abstracts/_config.scss#L41

https://github.com/Automattic/themes-workspace/blob/1d69ab9eadc0c8e809c7f88263b506d70fdf31f7/varya/assets/sass/abstracts/_config.scss#L45

Anecdotally, I've stumbled with how to use the foreground variables other than to color text. Would the intention be also for something like a border color?

PR #59, opened 2020-04-09
Varya: Removing light and dark border-color and variables.

allancole:
This PR removes the `--global--color-border-light` and `--global--color-border-dark` variables from the system and replaces each with the singular `--global--color-border` color variable. 

There were only a few instances of these colors in the theme already. 
- The comment-list border color
- A couple of minor instances of border colors on images in WooCommerce.

Fixes: #57

### DIFF
--- a/varya/assets/css/style-woocommerce-rtl.css
+++ b/varya/assets/css/style-woocommerce-rtl.css
@@ -1822,7 +1822,7 @@ body[class*="woocommerce"] #page .woocommerce-ordering select {
 }
 
 .single-product #page #woocommerce-wrapper div.product div.images .woocommerce-product-gallery__image--placeholder {
-	border-color: var(--global--color-border-light);
+	border-color: var(--global--color-border);
 }
 
 .single-product #page #woocommerce-wrapper div.product div.images .woocommerce-product-gallery__trigger {
@@ -1831,11 +1831,11 @@ body[class*="woocommerce"] #page .woocommerce-ordering select {
 }
 
 .single-product #page #woocommerce-wrapper div.product div.images .woocommerce-product-gallery__trigger::before {
-	border-color: var(--global--color-border-dark);
+	border-color: var(--global--color-border);
 }
 
 .single-product #page #woocommerce-wrapper div.product div.images .woocommerce-product-gallery__trigger::after {
-	background-color: var(--global--color-border-dark);
+	background-color: var(--global--color-border);
 }
 
 .single-product #page #woocommerce-wrapper div.product div.summary {
--- a/varya/assets/css/style-woocommerce.css
+++ b/varya/assets/css/style-woocommerce.css
@@ -1822,7 +1822,7 @@ body[class*="woocommerce"] #page .woocommerce-ordering select {
 }
 
 .single-product #page #woocommerce-wrapper div.product div.images .woocommerce-product-gallery__image--placeholder {
-	border-color: var(--global--color-border-light);
+	border-color: var(--global--color-border);
 }
 
 .single-product #page #woocommerce-wrapper div.product div.images .woocommerce-product-gallery__trigger {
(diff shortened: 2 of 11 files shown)


kjellr:
A couple things: 

1. I'd like to keep that comment border color the same — this PR changes it to a gray that's too dark. Do you think we should keep a `light` version here for that? Or should we auto-generate it somehow?
2. Also, these light/dark variables seem excessive to me too — I'm not sure folks will want to get this granular for each theme. Do we need these ones?

```
--global--color-foreground: #333333;
--global--color-foreground-light: #444444;
--global--color-foreground-dark: #000000;
--global--color-background: #FFFFFF;
--global--color-background-light: #FAFBF6;
--global--color-background-dark: #DDDDDD;
```

allancole:
> 1. I'd like to keep that comment border color the same — this PR changes it to a gray that's too dark. Do you think we should keep a light version here for that? Or should we auto-generate it somehow? (I'd lean towards keeping a light version in here.

@kjell, I think this is just a matter of choosing the right border-color from your comps (which I forgot to do in this PR). Unless you anticipate using multiple border colors as in **globally**, We can just change the default to the lighter color which I agree looks much nicer. 

> 2. Also, these light/dark variables seem excessive to me too — I'm not sure folks will want to get this granular for each theme. Do we need these ones?

Ahhh, this actually reminds me of something I completely forgot about. 

The `-light` and `-dark` modifiers for text-color and background are meant to be used together like this:

```
// Default
body {
   color: var(--global--color-foreground);
   background: var(--global--color-background);
}

// Light section
section.light {
   color: var(--global--color-foreground-light);
   background: var(--global--color-background-light);
}

// Dark section
section.dark {
   color: var(--global--color-foreground-dark);
   background: var(--global--color-background-dark);
}
```

This doesn’t come into play much for the “Botany” style. The default background is already perfect white so the `-light` modifiers should be the same values as the defaults since there is no _lighter_ color than white. The `-dark` modifiers work the same way but in the other contrast direction. Make sense? 

For a theme like “Skate Park” where the background is a bold green color, the `-light` modifiers come in handy if I wanted to add a section (Group Block, etc) with a lighter green background (from the Editor color palette) and have a predefined a11y-friendly lighter text-color already set for it. 

I’m not sure how well this part of the system even works since I’m only just now remembering that I did it this way for that reason :-). The way the color variables are named doesn’t make this approach obvious either which means we can probably get rid of it in favor of something more easy to digest. 

I do think a lighter text-color modifier can be kept though since those are often good for captions and other low-focus texts. I’m thinking we can change the name of the modifier to something more intuitive like `-dim`: 

```
--global--color-foreground: #333333;
--global--color-foreground-dim: #444444; 
--global--color-background: #FFFFFF;
```

kjellr:
> @kjell, I think this is just a matter of choosing the right border-color from your comps (which I forgot to do in this PR). Unless you anticipate using multiple border colors as in globally,

Yeah, in the case of this theme, I do want two different border colors (or at least, I think I do...). The default one is fine for occasional dividers, but it gets more frequent, close-together ones like the ones in the comments. I can probably just adjust things to use one border color.

kjellr:
> For a theme like “Skate Park” where the background is a bold green color, the -light modifiers come in handy if I wanted to add a section (Group Block, etc) with a lighter green background (from the Editor color palette) and have a predefined a11y-friendly lighter text-color already set for it.
> 
> The way the color variables are named doesn’t make this approach obvious either which means we can probably get rid of it in favor of something more easy to digest.

Do you think we could remove this all for now, and then add only when needed in child themes? Either that, or it seems like something that could be mostly automated by default.

allancole:
> Do you think we could remove this all for now, and then add only when needed in child themes? Either that, or it seems like something that could be mostly automated by default.

Yup i think we can just remove them for now. Then, if it’s something that we end up doing regularly in child themes, then we can always just add it back to the system as an enhancement.

allancole:
> Yeah, in the case of this theme, I do want two different border colors (or at least, I think I do...). The default one is fine for occasional dividers, but it gets more frequent, close-together ones like the ones in the comments. I can probably just adjust things to use one border color.

Yeah, my thinking on this is that we have just one _global_ border colors, for separators and comment borders like you mentioned. There are still other border-colors in the system that are element/block/component-specific like `--form--border-color`, `--pullquote--border-color`, `--quote--border-color`, etc. You can always set those individual variables to different colors on a per-theme basis, but I don’t think we’ll use multiple _global_ border color variants enough for them to be a useful part of the system.

PR #63, opened 2020-04-09
Varya: Correct menu animation jump when WooCommerce and Admin Bar are active

kjellr:
This is a weird one. Currently, if you have the admin bar and WooCommerce active, toggling the mobile menu or the mobile cart will cause the other button to jump down abruptly. Here's a slowed-down GIF so you can see this in action: 

menu-broken

This PR corrects that by adding a negative position to the button that's not being activated. This seems weird, but it works? 🤷 If anyone has a better explanation + fix for this, I'd love to hear it, but in the meantime, here's how it looks with this fix: 

menu-fixed

### DIFF
--- a/varya/assets/css/style-woocommerce-rtl.css
+++ b/varya/assets/css/style-woocommerce-rtl.css
@@ -722,14 +722,17 @@ body[class*="woocommerce"].wc-navigation-open #page .main-navigation #toggle-car
 }
 
 @media only screen and (max-width: 481px) {
-	body[class*="woocommerce"].admin-bar #page .main-navigation {
-		top: 46px;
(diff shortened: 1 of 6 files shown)


allancole:
> If we use overflow: hidden instead of position: fixed, then the cart button absolute positions wouldn't need to be adjusted. 

I just tried this and unfortunately, the background still scrolls behind the open menu for me. 

This is definitely a weird one, but I think I might have a solution for it. Gonna fork this PR and do some testing ;-).

kjellr:
> I can open up a separate issue, or we can quick fix it here?

Let's open a separate issue for that. 👍

PR #65, opened 2020-04-10
Varya: Update social icons to match Gutenberg's social links block

kjellr:
Addresses [half](https://github.com/Automattic/themes-workspace/issues/41#issuecomment-608410689) of #41

This PR updates the bundled social logos to match the ones used in Gutenberg's social links block. 

This matches [the approach used](https://github.com/WordPress/twentytwenty/blob/6d0a5240af108d02f58ec797fd49b32458a4c698/classes/class-twentytwenty-svg-icons.php#L173-L256) in Twenty Twenty, and gives us a clear license to point to in our readme (This PR adds that license note to `style.css` for now). 
Screen Shot 2020-04-09 at 8 10 39 PM


varya/style-rtl.css
```diff
@@ -19,6 +19,9 @@ Underscores is distributed under the terms of the GNU GPL v2 or later.
 
 Normalizing styles have been helped along thanks to the fine work of
 Nicolas Gallagher and Jonathan Neal https://necolas.github.io/normalize.css/
+
+Social Icons License: GNU GPL v2 or later.
```
jffng, line 23: Out of curiosity, is this more of a pre-caution to include this comment, given that this is the same license as the theme itself?

kjellr: Yeah, the important bit is the source link below this line — being able to show where we got them from.

PR #70, opened 2020-04-10
Varya: Woo and primary menu overlap on hover  

jffng:
When hovering on the main menu that has sub-menu, the sub-menu conflicts with the WooCommerce menu stacking order. This PR lowers the stacking order of the Woo menu.

**Before**
Kapture 2020-04-10 at 13 13 33

**After**
Screen Shot 2020-04-10 at 3 00 31 PM

### DIFF
--- a/varya/assets/sass/components/header/_primary-navigation.scss
+++ b/varya/assets/sass/components/header/_primary-navigation.scss
@@ -31,6 +31,10 @@
  	  	transform: translateY(var(--global--spacing-vertical));
 	}
 
+	.woocommerce-menu-container {
+		z-index: 498;
+	}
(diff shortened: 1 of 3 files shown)


jffng:
Is there a reason that the mobile menu uses `& > div`, instead of specifically naming the classes (`.main-menu-container` and `woocommerce-menu-container`)? 

https://github.com/Automattic/themes-workspace/blob/a6c2adb2f18ce9cbb2c7af37430b202ee2029ad1/varya/assets/sass/components/header/_primary-navigation.scss#L16-L17

jffng:
Closing in favor of #71

PR #71, opened 2020-04-10
Varya: Alternative approach to correcting menu animation jump

allancole:
This is a slightly different approach to the solution in #63.

- It simplifies the Sass structure so both mobile buttons respond to the `.lock-scrolling` class the same exact way. This means all of menu logic can be in `_primary-navigation.scss` and then the `_mini-cart.scss` only has to have overrides for button positioning and minor design related styles.
- Cleans up the z-index rules so both menus can never be visible at the same time.
- Doesn’t cause the unrelated issue [mentioned here](https://github.com/Automattic/themes-workspace/pull/63#pullrequestreview-391557866) to happen.
- Simplifies some of the common nested Sass selectors in a few places.

Before|After
------------|------------
|

### DIFF
--- a/varya/assets/css/style-woocommerce-rtl.css
+++ b/varya/assets/css/style-woocommerce-rtl.css
@@ -754,6 +754,9 @@ body[class*="woocommerce"] #page .main-navigation .woocommerce-cart-link .woocom
 body[class*="woocommerce"] #page .main-navigation .woocommerce-cart-link .woocommerce-cart-count {
 	color: var(--wc--mini-cart--color-count);
 	font-weight: normal;
+	white-space: nowrap;
+	display: inline-block;
+	vertical-align: bottom;
 }
 
 body[class*="woocommerce"] #page .main-navigation .woocommerce-cart-link .svg-icon {
@@ -765,7 +768,8 @@ body[class*="woocommerce"] #page .main-navigation .woocommerce-cart-link .svg-ic
 
 @media only screen and (min-width: 482px) {
 	body[class*="woocommerce"] #page .main-navigation .woocommerce-cart-link {
-		display: inline-block;
+		display: flex;
+		flex-wrap: nowrap;
 	}
 }
 
--- a/varya/assets/css/style-woocommerce-rtl.css
+++ b/varya/assets/css/style-woocommerce-rtl.css
@@ -683,20 +683,12 @@ body[class*="woocommerce"] #page .wc-block-grid__product-add-to-cart .added_to_c
 	text-decoration: none;
 }
 
-body[class*="woocommerce"] #page .main-navigation #toggle-cart {
-	position: absolute;
-	display: inline-block;
+body[class*="woocommerce"] #page .main-navigation > #toggle-cart {
 	right: 0;
-	margin: 0;
-	background-color: transparent;
-	color: var(--primary-nav--color-link);
+	top: 0;
 }
 
-body[class*="woocommerce"] #page .main-navigation #toggle-cart:hover {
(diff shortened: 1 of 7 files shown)


allancole:
Interesting. 

I’m unable to replicate your screenshot here:
image

Here’s how that same area works for me on this PR:


The odd thing is the primary-menu is fully centered with the mini-cart below it on your screenshot. This line adds flexbox which should be preventing that issue from happening. Super weird:

https://github.com/Automattic/themes-workspace/blob/9fe115193a6ebe161b3b467f8885bd913be2aa0d/varya/assets/sass/components/header/_primary-navigation.scss#L92-L93

I do like the stacked/centered UI in your screenshot a bit better (sans the overlapping of course). Should I go ahead and add that change to this PR, @jffng?

kjellr:
I'm seeing some bugs in this PR. 

Currently in Safari, the cart appears to sit in the center of the screen, on top of the main navigation. This makes it impossible to open the main nav: 

menu

In Chrome, it generally works, except there's a horizontal scroll when the admin bar is active:

menu-chrome

Also, just a UX comment: I see that the "close" button is left-aligned for the Woo cart in this new version, but I do think I prefer having it right-aligned all the time. It seems like a more consistent thing to do.

allancole:
> Currently in Safari, the cart appears to sit in the center of the screen, on top of the main navigation. This makes it impossible to open the main nav:

Just added a fix for that issue. 

RPReplay_Final1586792757
(Ignore the fallback fonts, ngrok URL isn’t secure so the fonts can’t load over SSL).

--

> In Chrome, it generally works, except there's a horizontal scroll when the admin bar is active

Having trouble recreating this one. @kjellr, can you double check to see that it’s the menu causing this. Here’s how that looks for me in chrome:



I wonder if its the admin bar thats actually causing the horizontal scroll?

allancole:
> Also, just a UX comment: I see that the "close" button is left-aligned for the Woo cart in this new version, but I do think I prefer having it right-aligned all the time. It seems like a more consistent thing to do.

Good point @kjellr! I do agree that its pretty common to have a close button be on the top right in these kinds of overlay UIs. But it also means we’d need one singular close button and 2 open buttons in the markup to do it properly. This would also make it way easier to design other kinds of header layouts in child themes, while keeping the standard menu overlay behavior.

The only issue is that if we decide to move the close button to always be on the right, I would want to refactor the markup/JS to optimize it for that specific behavior. Right now there’s only 2 buttons and they both work by toggling back and forth. If the Close button always appears on the right, it should be it’s own separate button, instead of an open/close toggle like it is now. 

Ideally this should be done in a separate, new menu PR.

kjellr:
> The only issue is that if we decide to move the close button to always be on the right, I would want to refactor the markup/JS to optimize it for that specific behavior. Right now there’s only 2 buttons and they both work by toggling back and forth. If the Close button always appears on the right, it should be it’s own separate button, instead of an open/close toggle like it is now.

Since the close button was on the right before this PR, I think moving it to the left here is a regression — I'd prefer it if we can get it back over to the right in this PR, and then go in later and refactor to make it cleaner if we'd like to do that.

allancole:
> I'd prefer it if we can get it back over to the right in this PR, and then go in later and refactor to make it cleaner if we'd like to do that.

Done in 533140d @kjellr 👍

allancole:
Also wanna point out that is PR also fixes that z-index issue on Desktop you mentioned @jffng :

**After:**

jffng:
Thanks @allancole, I verified that this indeed fixes the overlap issue. Going to close my PR in favor of this one.

PR #75, opened 2020-04-13
Varya: Fix button styles

kjellr:
Plugin versions of Gutenberg change button markup inside of the editor. Previously it looked like this: 

```
<!-- wp:button -->
<div class="wp-block-button"><a class="wp-block-button__link">My button</a></div>
<!-- /wp:button -->
```

Now, the button markup looks like this: 

```
<!-- wp:button -->
<a class="wp-block-button wp-block-button__link">My button</a>
<!-- /wp:button -->
```

Notice that there's no wrapping div anymore, and that the `wp-block-button__link` class has been relocated up to the same level as `wp-block-button`. As a result, the theme's current button styles do not work get inherited correctly in the editor: 

Screen Shot 2020-04-13 at 10 23 05 AM

In order to fix this while still preserving compatibility with older versions of Gutenberg, we need to _only_ target `wp-block-button__link`. This PR makes that change, and gets the editor styles for these buttons working again: 

Screen Shot 2020-04-13 at 10 24 13 AM

---

To test, please ensure that you try this against the current plugin version of Gutenberg (7.8.1), and also with the plugin deactivated.

### DIFF
--- a/varya/assets/css/style-editor.css
+++ b/varya/assets/css/style-editor.css
@@ -421,7 +421,7 @@ object {
 	color: currentColor;
 }
 
-.wp-block-button .wp-block-button__link {
+.wp-block-button__link {
 	color: var(--button--color-text);
 	font-weight: var(--button--font-weight);
 	font-family: var(--button--font-family);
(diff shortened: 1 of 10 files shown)


kjellr:
Latest updates here work pretty well in most scenarios, but they have a pretty major issue in the front end: 

Screen Shot 2020-04-13 at 12 23 13 PM
To reproduce, create a button _without_ the Gutenberg plugin active. Then view those buttons on the front end. Then activate the plugin and refresh. Can someone dig in and see what's causing this?

jffng:
@kjellr I followed your steps to reproduce that front-end error.

After activating the plugin, it looks like the old button block markup is still in tact, which contains the `wp-block-button` wrapping div that was removed in [this PR](https://github.com/WordPress/gutenberg/pull/21266).

Gutenberg's [default button block styles](https://github.com/WordPress/gutenberg/blob/c3798f141028240312dc7e7d530ad82b38900151/packages/block-library/src/button/style.scss#L14-L25) are then misapplied to the old markup, causing the front-end error.

Side note: after activating the plugin and adding a _new_ button to the page, Gutenberg automatically updated all the buttons on that page to reflect the new markup, resolving the issue. This doesn't seem like a good solution though.

kjellr:
Testing against the latest commits in Gutenberg, I _think_ this one is good to go now. Can someone else double check?

jffng:
Looks good, I only found one issue with old buttons using the outline style on the front end:

Screen Shot 2020-04-17 at 12 01 38 PM

I pushed a commit to remove that second border and verified that it worked with old and new buttons. 

It feels weird to set `border: none` [here](https://github.com/Automattic/themes-workspace/pull/75/commits/3ed2cf30edca56ac3f1810d82b2a0c370f7f3ee8#diff-0acd86d076deeea3c1e5feeddc55870dR1232-R1234). But finding the simplest (i.e. fewest lines of CSS) way to support both the old and new markup would take a deeper dive / refactor of the button styles.

kjellr:
@jffng according to https://github.com/WordPress/gutenberg/issues/21747, I think that might be a wider issue. Do you think we should push this for now, or just wait for that to resolved upstream?

jffng:
I think it's fine to push now. If / when that's fixed, the `border: none` rule would become unnecessary, but I don't think it'd break our styles.

kjellr:
Sounds good. 👍

Issue #74, opened 2020-04-13
Varya: Add Readme.txt


kjellr:
Varya is currently missing a readme.txt. We can find an example file here: 
https://make.wordpress.org/themes/handbook/review/required/#readme-txt-file

PR #80, opened 2020-04-13
Varya: Add initial readme.txt to pass TRT requirements.

allancole:
This is a first pass at adding a readme.txt file to pass the TRT requirements. I mostly just borrowed copy from the markdown version of this file, but let’s discuss what should actually go in here below. 

Fixes: #74

### DIFF
--- a/varya/assets/sass/style.scss
+++ b/varya/assets/sass/style.scss
@@ -11,11 +11,23 @@ License URI: LICENSE
 Text Domain: varya
 Tags: one-column, flexible-header, accessibility-ready, custom-colors, custom-menu, custom-logo, editor-style, featured-images, footer-widgets, rtl-language-support, sticky-post, threaded-comments, translation-ready
 
-This theme, like WordPress, is licensed under the GPL.
-Use it to make something cool, have fun, and share what you've learned with others.
+Varya WordPress Theme, (C) 2020 Automattic, Inc.
+Varya is distributed under the terms of the GNU GPL.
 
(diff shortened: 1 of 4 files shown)


kjellr:
I pushed a few small updates — mostly to the description and credit lines. I also synced these up with the `style.css` file. I think this is probably good to go. 

Quick question: for the copyright, do we need to credit anything specifically to Twenty Nineteen or Twenty Twenty?

allancole:
> do we need to credit anything specifically to Twenty Nineteen or Twenty Twenty?

Hmmm. I wanna say yes to Twenty Nineteen, but for Twenty Twenty, I’m not so sure. I’m sure there’s something in here from Twenty Twenty but I don’t know what exactly.

kjellr:
There's a function from Twenty Twenty in the colors PR. I'll at least call that out. Will submit a few other changes here later today. 👍

kjellr:
Ok, I've added some new updates to the license. Primarily, I've: 

- Noted that this is derivative of Twenty Nineteen. Based on the amount of little Twenty Nineteen code bits in here, I think that's true, right @allancole? 
- I've also noted that one of the functions used in #113 is straight from Twenty Twenty, since I think that's going to make it in. 

This should be good to go. 👍

allancole:
> this is derivative of Twenty Nineteen. Based on the amount of little Twenty Nineteen code bits in here, I think that's true, right @allancole?

Yeah, I’d say that’s pretty accurate and that this PR is good to go. 

LGTM! 👍

PR #82, opened 2020-04-14
Varya: Update selection color

### LINKED ISSUE #78: Varya: Text selection color might be too light.
The text selection color isn’t very noticeable at the moment. It currently matches the `--global--foreground-light` variable but I think a lighter version of the salmon color might be better.

Current screenshot:

image

Suggested alternative: 

image

kjellr:
Closes #78 

Before: 

Screen Shot 2020-04-14 at 8 09 45 AM

After: 

Screen Shot 2020-04-14 at 8 09 28 AM

### DIFF
--- a/varya/assets/sass/abstracts/_config.scss
+++ b/varya/assets/sass/abstracts/_config.scss
@@ -44,7 +44,7 @@ $typescale-ratio: 1.2; // Run ratio math on 1em == $typescale-base * $typescale-
 	--global--color-background-light: #FAFBF6;
 	--global--color-background-dark: #DDDDDD;
 	--global--color-border: #757575;
-	--global--color-text-selection: rgba( 163, 98, 101, 0.1 ); // --global--color-secondary in RGB, at 10% opacity
+	--global--color-text-selection: #F4EFEF;
 	--global--color-alert-success: yellowgreen;
 	--global--color-alert-info: skyblue;
 	--global--color-alert-warning: gold;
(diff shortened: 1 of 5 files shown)


varya/assets/sass/abstracts/_config.scss
```diff
@@ -44,7 +44,7 @@ $typescale-ratio: 1.2; // Run ratio math on 1em == $typescale-base * $typescale-
 	--global--color-background-light: #FAFBF6;
 	--global--color-background-dark: #DDDDDD;
 	--global--color-border: #757575;
-	--global--color-text-selection: #FAFBF6;
+	--global--color-text-selection: rgba( 163, 98, 101, 0.1 ); // --global--color-secondary in RGB, at 10% opacity
```
kjellr, line 47: I would've preferred to automate this in some way, e.g.: 

```
--global--color-text-selection: rgba( --global--color-secondary, 0.1 );
```

... however this only works if you define the original color variable via an RGB value, not Hex. We could do that here, but I worry that a child them might define the color in Hex, think it's working fine, but then be confused when there's no selection color at all. 😕 So I think this is the best route for now.

jffng: That could be an interesting idea to pose to the global styles crew. Since the global styles mechanism introduces a processing step (parsing the JSON file), certain color values could be computed / generated.

allancole: This is super interesting, but I think your right about possible confusion. There might be some PostCSS thing we could do to convert Hex to RGBa for everything but that’d probably be overkill. 

Also agreed that this would be a really clever thing for global styles to handle 👍.

allancole:
Oh wow! Yeah, I was wondering how you all achieved that border blocking trick. 

It’s not a deal breaker for me, but fixing it would probably mean not using the RGBa color in favor of a Hex color which I’d probably be okay with too.

kjellr:
Ah, interesting. So we can fix that by just switching to a hex color? Maybe should just do that in this case then.

allancole:
Yup, standard Hex color will probably best. 👍

kjellr:
This one should be ready to go!

PR #83, opened 2020-04-14
Varya: Refine mobile heading spacing

kjellr:
This PR adjusts the spacing of the site header on small screens. It adds a little space to ensure that the whitespace balance is a little more in line with the way it is on desktop, and it also introduces a little extra top margin if a main navigation exists, so that the logo and tagline appear more centered vertically.

Before: 

Menu|No Menu
----|----
dotorgthemes test_ (2)|dotorgthemes test_ (3)

After:

Menu|No Menu
----|----
dotorgthemes test_|dotorgthemes test_ (1)

### DIFF
--- a/varya/assets/sass/structure/_vertical-margins.scss
+++ b/varya/assets/sass/structure/_vertical-margins.scss
@@ -27,12 +27,20 @@
 }
 
 .site-header {
-	padding-top: calc(2 * var(--global--spacing-vertical));
-	padding-bottom: calc(2 * var(--global--spacing-vertical));
+	padding-top: calc(3 * var(--global--spacing-vertical));
+	padding-bottom: calc(3 * var(--global--spacing-vertical));
+
+	&.has-main-navigation {
+		padding-top: calc(4 * var(--global--spacing-vertical));
+	}
 
 	@include media(mobile) {
 		padding-top: calc(3 * var(--global--spacing-vertical));
 		padding-bottom: calc(3 * var(--global--spacing-vertical));
+
+		&.has-main-navigation {
+			padding-top: calc(3 * var(--global--spacing-vertical));
+		}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


varya/header.php
```diff
@@ -23,7 +23,7 @@
 <div id="page" class="site">
 	<a class="skip-link screen-reader-text" href="#content"><?php _e( 'Skip to content', 'varya' ); ?></a>
 
-		<header id="masthead" class="site-header default-max-width">
+		<header id="masthead" class="site-header default-max-width <?php if ( has_nav_menu( 'primary' ) ) : echo esc_attr( 'has-main-navigation' ); endif; ?>">
```
allancole, line 26: I think this class would be better on the `<body>` since we already have a `body_class` filter for this kind of thing:

https://github.com/Automattic/themes-workspace/blob/d01dc09e8bffd538514dcf2e713c1d78be34c1f8/varya/inc/template-functions.php#L28-L45

That filter is one of those long standing best practices from `_s` that many Themers are already familiar with.

varya/assets/sass/structure/_vertical-margins.scss
```diff
@@ -27,12 +27,20 @@
 }
 
 .site-header {
-	padding-top: calc(2 * var(--global--spacing-vertical));
-	padding-bottom: calc(2 * var(--global--spacing-vertical));
+	padding-top: calc(3 * var(--global--spacing-vertical));
+	padding-bottom: calc(3 * var(--global--spacing-vertical));
+
+	&.has-main-navigation {
```
allancole, line 33: Moving `.has-main-navigation` to `<body>` means we’ll have to change these selectors to something like this. 

`.has-main-navigation & {}`

varya/header.php
allancole: Lines 38-42 are obsolete and can be removed entirely (along with the image_filter function itself).

kjellr: Ah, excellent. Good call. 👍

kjellr:
@allancole the latest commits should take care of your suggestions. Mind giving this another review?

PR #86, opened 2020-04-15
Use the theme's max-width in the editor

kjellr:
Before: 

Screen Shot 2020-04-15 at 2 18 19 PM

After:

Screen Shot 2020-04-15 at 2 29 10 PM

### DIFF
--- a/varya/assets/css/style-editor.css
+++ b/varya/assets/css/style-editor.css
@@ -121,6 +121,10 @@ a {
 	max-width: var(--responsive--aligndefault-width);
 }
 
+.wp-block[data-align="wide"], .wp-block.alignwide {
+	max-width: var(--responsive--alignwide-width);
+}
+
 div[data-type="core/button"] {
 	display: block;
 }
--- a/varya/assets/sass/base/_editor.scss
+++ b/varya/assets/sass/base/_editor.scss
@@ -30,11 +30,18 @@ a {
 	cursor: pointer;
 }
 
-// Gutenberg injects a rule that limits the max width of .wp-block to 580px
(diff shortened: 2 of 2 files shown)


varya/assets/css/style-editor.css
```diff
@@ -121,6 +121,10 @@ a {
 	max-width: var(--responsive--aligndefault-width);
 }
 
+.wp-block[data-align="wide"], .wp-block.alignwide {
+	max-width: var(--responsive--alignwide-width);
```
kjellr, line 125: I opted to use this variable instead of `@extend %responsive-alignwide-width;` here because we do not need to apply `margin-left` and `margin-right` rules.

varya/assets/sass/base/_editor.scss
```diff
@@ -30,12 +30,19 @@ a {
...
+	// Gutenberg injects a rule that limits the max width of .wp-block to 580px
+	// This line overrides it to use the responsive spacing rules for default width content
 	&:not([data-align="full"]):not([data-align="wide"]){
 		max-width: var(--responsive--aligndefault-width);
 	}
+
+	// Use the theme's max-width for wide alignment.
+	&[data-align="wide"],
+	&.alignwide {
+		max-width: var(--responsive--alignwide-width);
```
kjellr, line 44: I opted to use this variable instead of `@extend %responsive-alignwide-width;` here because we do not need to apply `margin-left` and `margin-right` rules.

allancole: Ahhh, clever! Happy to see it was easy to sort out… thought I might need to write another responsive logic system just for the editor 👍

Issue #89, opened 2020-04-15
Varya: 404 page treatment


jffng:
Currently the 404 page displays the following:

<img width="687" alt="Screen Shot 2020-04-15 at 3 46 47 PM" src="https://user-images.githubusercontent.com/5375500/79381708-6487ca80-7f30-11ea-8503-2b3e28d92156.png">

I think it's okay, but we may want to consider dropping the search input if it's already at the top of the footer.

allancole:
That second search form is likely coming from the footer widget area. If users don’t have a search form in the footer widgets, than they’ll see just one search form.

jffng:
Got it! 

Thoughts on adjusting the search template to use more of a flex layout like so?

<img width="700" alt="Screen Shot 2020-04-15 at 5 26 13 PM" src="https://user-images.githubusercontent.com/5375500/79390763-4f199d00-7f3e-11ea-8f65-7a0bfde594e6.png">

allancole:
Yup, I like that. Go for it! 👍

PR #91, opened 2020-04-16
Varya: Adjust search form layout

jffng:
Adjusts the layout of the `.search-form`, which shows up on a 404 page right now or appears when you activate the search widget. #89 

**Before**
Screen Shot 2020-04-16 at 9 52 45 AM

**After**
Screen Shot 2020-04-16 at 9 53 01 AM

### DIFF
--- a/varya/assets/sass/elements/_forms.scss
+++ b/varya/assets/sass/elements/_forms.scss
@@ -41,3 +41,16 @@ input[type=checkbox] + label {
 	margin-left: 0.5em;
 	line-height: 1em;
 }
+
+.search-form {
+	display: flex;
+	margin-bottom: var(--global--spacing-vertical);
+	> label {
+		display: flex;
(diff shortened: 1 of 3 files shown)


varya/assets/sass/elements/_forms.scss
```diff
@@ -41,3 +41,16 @@ input[type=checkbox] + label {
 	margin-left: 0.5em;
 	line-height: 1em;
 }
+
+.search-form {
+	display: flex;
+	margin-bottom: var(--global--spacing-vertical);
```
allancole, line 47: Let’s remove the bottom margin here. Vertical spacing is already applied by the parent element in this file which covers both the widget and elements on the 404 page inside `.page-content`: 

https://github.com/Automattic/themes-workspace/blob/master/varya/assets/sass/structure/_vertical-margins.scss

Issue #93, opened 2020-04-16
Varya: Add gradient presets


kjellr:
We need to add some [custom gradient presets](https://developer.wordpress.org/block-editor/developers/themes/theme-support/#block-gradient-presets) into the theme. 

We should include some with abrupt transitions like this... 

![Screen Shot 2020-04-16 at 3 45 47 PM](https://user-images.githubusercontent.com/1202812/79499883-9c0e7980-7ff9-11ea-8860-20ee84d9c4ac.png)

... to be used to create this representation from the comps: 

![Screen Shot 2020-04-16 at 3 45 56 PM](https://user-images.githubusercontent.com/1202812/79499899-a3358780-7ff9-11ea-962d-ad0271d6f5de.png)

jffng:
Closed by #95.

PR #95, opened 2020-04-16
Varya: try adding gradient preset

jffng:
This PR adds a single gradient preset (#93) implementing the diagonal cutting background: 

Screen Shot 2020-04-16 at 5 11 04 PM

A few considerations:
- Where should something like this live in Varya? For now I threw it into a new file under `sass/base` called `_presets.scss` cc @allancole 
- Child themes can easily take advantage of this preset by quickly changing the colors via a CSS variable change. However, the implementation of custom gradient presets require that you declare them in the `functions.php` file, so the editor can show a preview in the block settings. This is just a bit more maintenance to keep it in sync, but I'm noting this so that our child theme build tool can account for this it, unless there's another approach.
- There's some slight aliasing happening along the diagonal. Is this acceptable?

### DIFF
--- a/varya/assets/css/style-editor.css
+++ b/varya/assets/css/style-editor.css
@@ -90,7 +90,7 @@
  * Output
  */
 /* Customizations afforded by the Gutenberg editor */
-.has-red-to-cream-gradient-background {
+.has-hard-diagonal-gradient-background {
 	background: linear-gradient(to bottom right, var(--global--color-secondary) 50%, var(--global--color-background-light) 50%);
 }
 
--- a/varya/assets/css/style-editor.css
+++ b/varya/assets/css/style-editor.css
@@ -1124,7 +1124,35 @@ table th,
  * Custom gradients 
 */
 .has-hard-diagonal-gradient-background {
-	background: linear-gradient(to bottom right, var(--global--color-secondary) 49.9%, var(--global--color-background-light) 50%);
+	background: linear-gradient(to bottom right, #A36265 49.9%, #FAFBF6 50%);
+}
+
+.has-hard-diagonal-inverted-gradient-background {
+	background: linear-gradient(to top left, #A36265 49.9%, #FAFBF6 50%);
+}
+
+.has-diagonal-gradient-background {
+	background: linear-gradient(to bottom right, #A36265, #FAFBF6);
+}
+
+.has-diagonal-inverted-gradient-background {
+	background: linear-gradient(to top left, #A36265, #FAFBF6);
+}
+
+.has-hard-horizontal-gradient-background {
+	background: linear-gradient(to bottom, #A36265 50%, #FAFBF6 50%);
+}
+
+.has-hard-horizontal-inverted-gradient-background {
+	background: linear-gradient(to top, #A36265 50%, #FAFBF6 50%);
+}
+
+.has-horizontal-gradient-background {
+	background: linear-gradient(to bottom, #A36265, #FAFBF6);
+}
+
+.has-horizontal-inverted-gradient-background {
+	background: linear-gradient(to top, #A36265, #FAFBF6);
 }
 
 /**
--- a/varya/assets/css/style-editor.css
+++ b/varya/assets/css/style-editor.css
@@ -1124,7 +1124,7 @@ table th,
  * Custom gradients 
 */
 .has-hard-diagonal-gradient-background {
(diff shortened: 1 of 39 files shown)


jffng:
Ah, I'm an idiot and didn't finish reading the docs 😅 .

> Themes are responsible for creating the classes that apply the gradients. 

Pushing that shortly.

allancole:
> Where should something like this live in Varya? For now I threw it into a new file under sass/base called _presets.scss cc @allancole

Feels like that should go in `/varya/assets/sass/blocks/utilities/_presets.scss` 

> There's some slight aliasing happening along the diagonal. Is this acceptable?

Hmmm, is that a common issue with diagonals?

kjellr:
> There's some slight aliasing happening along the diagonal. Is this acceptable?

Mind adding a screenshot?

jffng:
Aliasing along the diagonal screenshot:

Screen Shot 2020-04-17 at 9 39 16 AM

kjellr:
I wonder if @jasmussen or @MichaelArestad happens to know anything about smoothing out that messy gradient aliasing? It feels like there's gotta be some sort of hack out there, but I haven't found it yet...

Screen Shot 2020-04-21 at 1 23 37 PM

I tried applying slight blur via something like [`backdrop-filter`](https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter), but it gets too blurry.

jasmussen:
I'm mobile at the moment so I haven't been able to test. But looking at the code it appears one half if the diagonal split stops at 50%, and the next starts at 50%. The aliasing between those two colors might simply by the entire color range being compressed into a singularity.

So my first instinct would be to try and end the first half of the gradient at 49.9%, and start the second half of the gradient at 50.1%, and seeing if allowing a tiny bit of gradation could actually translate into a form of antialiasing. And if yes, I'd massage those percentages to be as close as possible.

kjellr:
> So my first instinct would be to try and end the first half of the gradient at 49.9%, and start the second half of the gradient at 50.1%, and seeing if allowing a tiny bit of gradation could actually translate into a form of antialiasing. And if yes, I'd massage those percentages to be as close as possible.

Good thought! Unfortunately, I just tried it out but it appears to have no effect. 😞 

I tried also adding a middle 1% of an average color (the average of those two), but it didn't seem to fix anything either.

jasmussen:
Thanks for trying. Bummer.

This answer suggests there might be some fiddling that can be done:

https://stackoverflow.com/a/33094994

jffng:
I tested @jasmussen's idea, and it helped create some antialiasing, I _think_.

50:50 | 49.9:50
--- | ---
Screen Shot 2020-04-21 at 2 10 14 PM|Screen Shot 2020-04-21 at 2 10 04 PM

jasmussen:
Oh wow yeah that does look slightly better! I'm sure it's a balance to get just right but the pixels appear gone.

kjellr:
Oh weird! I wonder why I didn't see that on my end when I tried it... this looks way better. 🎉

allancole:
Yup, that looks much better now, Jeff 👍

jffng:
> I wonder why I didn't see that on my end when I tried it

After some more testing, it only improved when the darker color stopped earlier (i.e. 49.9%), not the other way around. 

I just pushed what I thought created the best balance, and tested it across backgrounds of varying width and height. There is one use case I came across that it doesn't scale to very well: a narrow height.

Screen Shot 2020-04-22 at 9 18 31 AM

Screen Shot 2020-04-22 at 9 21 00 AM

When I try to account for this, it degrades the backgrounds at other scales. I don't think it's a dealbreaker, as most everything else looks good. Can we merge, try using it, and adjust as needed?

kjellr:
> When I try to account for this, it degrades the backgrounds at other scales. I don't think it's a dealbreaker, as most everything else looks good. 

Yeah, I think that's okay. Users will manually be selecting this gradient, so if it looks bad, they can just change this. 

> Can we merge, try using it, and adjust as needed?

Since we've already got this PR going, can we add just a few more so we've got all our bases covered? Some ideas: 

Gradients

jffng:
@kjellr I added those presets and encountered a bug in Gutenberg, filed here https://github.com/WordPress/gutenberg/issues/21831. 

It only happens with the horizontal hard break preset.

allancole:
Yup, this is good to go. 
Rebase against master to fix that conflict and then LGTM! 👍

PR #108, opened 2020-04-16
Add tiny font size styles

### LINKED ISSUE #100: Varya: Tiny font preset is not styled correctly in the editor or front end
Screen Shot 2020-04-16 at 6 10 43 PM

To reproduce: 

1. Add a paragraph block
2. Use the sidebar to choose the "Tiny" [font preset](https://cloudup.com/cUg0KHgHP6A).
3. Note that the text should be 14px, but instead it appears to be 18px.

kjellr:
Fixes #100 

Before: 

Screen Shot 2020-04-16 at 7 03 00 PM

After: 

Screen Shot 2020-04-16 at 7 03 11 PM

### DIFF
--- a/varya/assets/css/style-editor.css
+++ b/varya/assets/css/style-editor.css
@@ -1076,6 +1076,11 @@ table th,
 	color: var(--global--color-primary);
 }
 
+.is-tiny-text,
+.has-tiny-font-size {
+	font-size: var(--global--font-size-xs);
+}
+
 .is-small-text,
 .has-small-font-size {
 	font-size: var(--global--font-size-sm);
--- a/varya/assets/sass/blocks/utilities/_editor.scss
+++ b/varya/assets/sass/blocks/utilities/_editor.scss
@@ -111,6 +111,11 @@
 }
 
 // Gutenberg Font-size utility classes
+.is-tiny-text,
+.has-tiny-font-size {
+	font-size: var(--global--font-size-xs);
+}
+
 .is-small-text,
 .has-small-font-size {
 	font-size: var(--global--font-size-sm);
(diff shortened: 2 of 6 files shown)


kjellr:
Note that front-end styles for most of these styles are currently broken because they appear to be overridden by some `:root` styles. I think those come from Gutenberg? We should try to fix that in this PR too I suppose. 

Screen Shot 2020-04-16 at 7 04 19 PM

Screen Shot 2020-04-16 at 7 06 39 PM

allancole:
Yeah, this is definitely a Gutenberg / global-styles issue. 

The value thats coming from Gutenberg for `:root .has-large-font-size { font-size: 36px; }` should match the theme’s font size settings in `functions.php`. 

These lines should set that default large font-size to `20px` but it doesn’t: 
https://github.com/Automattic/themes-workspace/blob/30438ed11546b6e2b20c8198fbf1eafc5fb14413/varya/functions.php#L138-L143

allancole:
One way to fix it on our end would be to add `:root` to the beginning all of our font-size utility classes:

I tried a test on this PR and it works: 
image

However, we should investigate and figure out if this is something that should be solved in Gutenberg first before fixing it on our end.

kjellr:
> we should investigate and figure out if this is something that should be solved in Gutenberg first before fixing it on our end.

Agreed — that level of specificity seems weird. Would you mind opening an issue there?

allancole:
Filed an issue for this here: https://github.com/WordPress/gutenberg/issues/21759

jffng:
Since @allancole opened an issue, I think we should just go forward targeting `:root` for now. It feels overly specific, but it won't break if Gutenberg fixes it.

PR #110, opened 2020-04-17
Varya: Use theme fonts for initial paragraph inserter

### LINKED ISSUE #98: Varya: Block placeholder does not inherit correct font specs on first load.
Screen Shot 2020-04-16 at 5 48 15 PM

^ "Start writing or type..." is using Noto Sans, not our theme's font.

jffng:
Fixes #98. 

I put this in `sass/base/_editor.scss` because it feels related to the base styles of the editor, since this only happens on initial render. 

**Before**
Screen Shot 2020-04-17 at 11 38 29 AM

**After**
Screen Shot 2020-04-17 at 11 32 34 AM

### DIFF
--- a/varya/assets/sass/base/_editor.scss
+++ b/varya/assets/sass/base/_editor.scss
@@ -43,6 +43,11 @@ a {
 	&.alignwide {
 		max-width: var(--responsive--alignwide-width);
 	}
+
+	&.block-editor-default-block-appender > textarea {
+		font-family: var(--global--font-secondary);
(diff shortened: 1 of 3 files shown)


varya/assets/sass/base/_editor.scss
```diff
@@ -43,6 +43,11 @@ a {
 	&.alignwide {
 		max-width: var(--responsive--alignwide-width);
 	}
+
```
allancole, line 46: Yup, this is the perfect place for it.

Issue #87, opened 2020-04-15
Varya: pagination treatment


jffng:
If the front page is set to display the latest posts, the pagination looks like this:

<img width="646" alt="Screen Shot 2020-04-15 at 3 25 57 PM" src="https://user-images.githubusercontent.com/5375500/79379929-acf1b900-7f2d-11ea-82ad-b53031e17dec.png">

I'm raising an issue in case we'd like a different treatment since the theme doesn't use bold typography anywhere else.

allancole:
Good catch! Changing the font to the san-serif would fix this for me, but I differ to @kjellr for design direction ;-)

jffng:
Quick test just removing the font-weight:

<img width="683" alt="Screen Shot 2020-04-16 at 10 18 56 AM" src="https://user-images.githubusercontent.com/5375500/79467407-bf223480-7fcb-11ea-9c6b-f00aa82c18dc.png">

Using sans-serif and `--global--font-size-sm`:
<img width="728" alt="Screen Shot 2020-04-16 at 10 17 26 AM" src="https://user-images.githubusercontent.com/5375500/79467169-81251080-7fcb-11ea-9482-151c1af03501.png">

One other consideration -- there is a class for the currently selected page. I just added a bottom border to show that it's there:
<img width="661" alt="Screen Shot 2020-04-16 at 10 15 54 AM" src="https://user-images.githubusercontent.com/5375500/79467004-4c18be00-7fcb-11ea-9ee7-9d456e961060.png">

allancole:
Ohhh nice, I like that bottom-border! I like the normal weight serif too, maybe we can combine the two? The underline on the serif matches other similar design decisions in post headings but I like how the grey border-color makes it more of a visual indicator and less of a link 👍

kjellr:
I think since that's UI, it makes sense for it to be treated with the same font styles as our post meta (sans serif, either `--global--font-size-sm` or `--global--font-size-xs`). I also like the underline there. 👍

PR #111, opened 2020-04-17
Varya: update pagination treatment and utilize existing pagination variables

jffng:
Addresses #87.

I discovered there were unused pagination variables. I'm not sure if these are warranted, but made use of them anyways. cc @allancole in case you were thinking of removing these in the audit. 

**Before**
Screen Shot 2020-04-17 at 1 13 00 PM

**After**
Screen Shot 2020-04-17 at 1 08 45 PM

### DIFF
--- a/varya/assets/sass/variables.scss
+++ b/varya/assets/sass/variables.scss
@@ -30,6 +30,7 @@
 	@include entry-variables();
 	@include comments-variables();
 	@include footer-variables();
+	@include pagination-variables();
 	/* Vendors */
 	@include woocommerce-variables();
 }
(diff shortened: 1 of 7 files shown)


varya/assets/sass/variables.scss
```diff
@@ -30,6 +30,7 @@
 	@include entry-variables();
 	@include comments-variables();
 	@include footer-variables();
+	@include pagination-variables();
```
allancole, line 33: Didn’t realize this was missing—good catch!

kjellr:
Looks great, thanks @jffng!

Issue #101, opened 2020-04-16
Varya: Search widget block styles need a little cleanup


kjellr:
- [x] On the frontend, the "Search" label should not be bold: 

![Screen Shot 2020-04-16 at 6 22 46 PM](https://user-images.githubusercontent.com/1202812/79512426-6aa0a880-800f-11ea-8461-0fc1a672b015.png)

- [x] In the editor, the button and form styles should match the front end: 

![Screen Shot 2020-04-16 at 6 23 06 PM](https://user-images.githubusercontent.com/1202812/79512481-81df9600-800f-11ea-83e4-ebe4d3dc1765.png)

PR #112, opened 2020-04-17
Varya: Refine the search block styles

jffng:
Addresses #101.

**Before**
Editor | Front End
------------|------------
Screen Shot 2020-04-17 at 4 06 56 PM|Screen Shot 2020-04-17 at 4 07 03 PM

**After**
Editor | Front End
------------|------------
Screen Shot 2020-04-17 at 4 06 03 PM|Screen Shot 2020-04-17 at 4 00 00 PM

(h/t @allancole for this cool table markdown formatting)

Some considerations:
- Previously there were no search block styles and all the styling took from `elements/_forms.scss`. I felt it would be better to include block styles for this rather than put them in `_forms.scss`, but it makes a handful of styles redundant. I think this still may be the cleanest solution to support both the widget and the search block. 
- I didn't include a new config for the search block, and borrowed the existing `form` variables.

### DIFF
--- a/varya/assets/sass/blocks/search/_style.scss
+++ b/varya/assets/sass/blocks/search/_style.scss
@@ -5,7 +5,17 @@
         font-weight: normal;
     }
     .wp-block-search__input {
+        border: var(--form--border-width) solid var(--form--border-color);
+        border-radius: var(--form--border-radius);
+        color: var(--form--color-text);
+        line-height: var(--global--line-height-body);
         max-width: inherit;
         margin-right: calc( .66 * var(--global--spacing-horizontal) );
+        padding: var(--form--spacing-unit);
+
+        &:focus {
+            color: var(--form--color-text);
+            border-color: var(--form--border-color);
+        }
     }
 }
--- a/varya/assets/sass/blocks/search/_style.scss
+++ b/varya/assets/sass/blocks/search/_style.scss
@@ -1,0 +1,11 @@
+.wp-block-search {
+    display: flex;
+    max-width: var(--responsive--aligndefault-width);
+    .wp-block-search__label {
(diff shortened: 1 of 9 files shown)


varya/assets/sass/blocks/search/_style.scss
```diff
@@ -0,0 +1,11 @@
+.wp-block-search {
+    display: flex;
+    max-width: var(--responsive--aligndefault-width);
```
jffng, line 3: Making an assumption here that the search block isn't going to go beyond wide or full widths.

allancole: Yeah, I think this assumption is perfectly fine.

varya/assets/sass/elements/_forms.scss
```diff
@@ -42,8 +42,12 @@ input[type=checkbox] + label {
 	line-height: 1em;
 }
 
+// Search widget styles
```
allancole, line 45: Lets move these search widget specific styles to `varya/assets/sass/components/widgets/_style.scss`

varya/assets/sass/elements/_forms.scss
```diff
@@ -13,12 +13,12 @@ input[type="time"],
 input[type="datetime"],
 input[type="datetime-local"],
 input[type="color"],
+input.wp-block-search__input,
```
allancole, line 16: I think this `input.wp-block-search__input` selector is okay here, but part of me wants this move to the Search Block Sass file itself. Might make it easier to find even if it does mean duplicating the input styles.

jffng:
Made those adjustments to the search block style locations, going to merge 🚢 .

PR #113, opened 2020-04-20
Varya: Add color options to the Customizer

### LINKED ISSUE #94: Varya: Build in color-editing capabilities.
We'll need to include a way for users to edit the colors used in the theme. Until Global Styles arrive, this should probably be done via the Customizer.

kjellr:
Will eventually close #94 

This PR begins to add custom color options to the Customizer. I realized I built this one morning last fall for the original Varia, so I just ported over the changes for the most part. Still has a few things that need to be fixed: 

- [x] The color variables defined here are global, but they shouldn't be. 
- [x] We shouldn't really be adding the custom CSS here to the `head`. It should be properly enqueued instead. 
- [x] We need to change the [Gutenberg custom colors](https://github.com/Automattic/themes-workspace/blob/30438ed11546b6e2b20c8198fbf1eafc5fb14413/varya/functions.php#L153-L183) to reflect these ones. 

Screenshot: 

custom-colors


kjellr:
I pushed a bunch of updates this morning. 

custom-colors

Still left to do: 

- [ ] Fix editor styles. I added these in 65a7322, and it _should_ work, but `wp_add_inline_style()` doesn't appear to be rendering any of the custom styles. 🤔 @allancole can you take a look, since you have much more customizer experience than I do?
- [ ] The [custom Gutenberg colors are working ](https://github.com/Automattic/themes-workspace/pull/113/commits/1b86db39566c64cbe2f65fd266a696c837c1d476)but they need to be adjusted to _only_ work if the "Custom" colors are active. Right now if you change to custom colors, modify the colors, and then switch back to the default, Gutenberg colors will still show the custom ones.  
- [ ] We need to auto-generate any additional variables (like hover or text selection colors) when you select custom ones.

allancole:
I think those last few commits have this in a pretty good spot and address 2 of the last 3 bullet points :-) 

image

There’s just one piece that still needs to be solved: 

> We need to auto-generate any additional variables (like hover or text selection colors) when you select custom ones.

The hardest part about this is we need the autogenerated colors to be aware of the background color so there’s always enough contrast to make hover and text-selection colors legible. Here’s a couple possible solutions I have in mind: 

1) We can use PHP to generate these colors, but then we’d need to bundle some kind of color pre-processor in the theme like what we did for Twenty Nineteen and the HSL color converter. I found one snippet that looks fairly promising here and does most of what we need: https://gist.github.com/stephenharris/5532899#gistcomment-2251604

2) The seemingly easiest way to solve this would be to just set the hover color to be the same as the foreground color whenever custom colors are in use. I imagine this would work well as long as the foreground and primary/secondary colors are different. This doesn’t address text selection colors at all though. We’d probably have to remove the text selection color and replace it with  `color: inherit` so that browsers can determine the text-selection color again.

kjellr:
> We can use PHP to generate these colors, but then we’d need to bundle some kind of color pre-processor in the theme like what we did for Twenty Nineteen and the HSL color converter. I found one snippet that looks fairly promising here and does most of what we need: https://gist.github.com/stephenharris/5532899#gistcomment-2251604

This seems like a decent, flexible approach. 👍

kjellr:
@allancole Check out c71dd4b — this is what I was thinking yesterday to avoid having two `varya_generate_custom_color_variables()` functions. It seems to work just fine in my testing.

varya/classes/class-varya-custom-colors.php
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+
+	/**
+	 * Generate color variables for customizer & frontend.
+	 */
+	function varya_generate_custom_color_variables( $context = null ) {
+
+		if ( $context === 'editor' ) {
+			$theme_css = ':root .editor-styles-wrapper {';
```
jffng, line 123: Is it worth inserting a comment here about why this is necessary? Asking because Gutenberg was trying to move away from having to target this class, IIRC.

jffng: On second thought, it's pretty straightforward from just reading the if statement.

allancole:
@kjellr yup, c71dd4b is spot on. So mad I didn’t even consider simply passing a variable through the function, haha. Good catch! 👍

allancole:
These last couple of commits add a way of auto-generating the colors following these rules:  
- `--global--color-primary-hover`: the midpoint color between Primary and Background
- `--global--color-secondary-hover`: the midpoint color between Secondary and Background
- `--global--color-foreground-light`: the midpoint color between Foreground and Background
- `--global--color-background-light`: the midpoint color between Background and Foreground

It still needs work, but for the most part this produces pretty solid and mostly a11y-friendly color schemes with ease: 

Default|Custom
------------|------------
image|varya-custom-colors

It isn’t perfect though and certain combinations don’t always look great a11y-wise. Might be worth it to add some sort of contrast test to the auto-generated colors, but that would also mean adding a lot more php code to do the color processing. 

I also removed the `--global--color-background-light` option from the customizer. I’m not sure its necessary to keep that one. Since it’s related to the background-color, I figured it’d be better to auto-generate it.

kjellr:
Awesome, I do think we need some adjusting to make this work well, but this is an excellent start. In general, I'm not sure a midpoint is actually the right measurement here. In most cases, the colors here seem much more light or dark than the original ones. Can we try to adjust them to be closer? I think our hover colors should be more subtle in general — I think we should auto-generate only subtle changes, and allow for users to manually adjust anything that's not going to be subtle. 

> Might be worth it to add some sort of contrast test to the auto-generated colors, but that would also mean adding a lot more php code to do the color processing.

In general, I'd like us to file a post-release issue to integrate some sort of a11y warnings in here. 
> I also removed the --global--color-background-light option from the customizer. I’m not sure its necessary to keep that one. Since it’s related to the background-color, I figured it’d be better to auto-generate it.

Just wanted to note that the way we're using it (and because of the way it can be applied via the editor) though, it's not necessarily related to background color:

background-light

Also: Is there any way we can keep the original colors _unless_ a user has manually changed a color from its default? So for example, if someone were to choose the "Custom" radio button, but then not touch any of the color controls, we'd still use the defaults for each. And if the user were to only adjust the Primary Color, then we'd only generate the `--global--color-primary-hover` value, and leave the rest as is.

jffng:
The work here is super cool 👏 , I've learned a ton from following the code and conversation.

> I'd like us to file a post-release issue to integrate some sort of a11y warnings in here

Are we comfortable shipping this without warning the user when they create an accessibility issue, say through a low-contrast background : primary color combination? 

I don't think it makes sense to give users a way to change the color through the customizer unless we integrate warnings or make inaccessible color contrast ratios impossible, and that's more work. Also keeping in mind that Gutenberg / global styles should eventually take care of this, granted we don't know how far away that is.

> Might be worth it to add some sort of contrast test to the auto-generated colors, but that would also mean adding a lot more php code to do the color processing.

Rather than (or in addition to) adding these features to the customizer, we could integrate some color processing utilities with the child theme build tools also.

allancole:
> In general, I'm not sure a midpoint is actually the right measurement here. In most cases, the colors here seem much more light or dark than the original ones. 

Yeah math-wise it was just easier to figure out the halfway point. Agreed that the colors should be more subtle, though. I’ll play around with the math to see if there’s a way to get closer to the default contrasts. 

> In general, I'd like us to file a post-release issue to integrate some sort of a11y warnings in here.

Yeah, that might be nice to have too, but shouldn’t be a blocker in my eyes. 

> Just wanted to note that the way we're using it (and because of the way it can be applied via the editor) though, it's not necessarily related to background color:

Ahh, I knew there was gonna be something I missed when I removed the background-light color from the options! This may need further discussion but I think because of how that _light-tan_ color is being using in this theme it _might_ be better to add a new `--global--color-tertiary` variable, rather than tying it to the background-color variant variables. That way it can becomes an editor palette color (solving the unexpected color change problem in Kjell’s screenshot) and allows child themes to use the variable however/wherever they want. When we assign it to a background-color, I think it can potentially make child theming confusing.

This is partly because the `-light` suffix is misleading. It should probably be renamed to `-dim` and here’s why. The background-color is always a singular fixed color. All other colors sit on top of it and are effectively _foreground_ colors (even though some are named, primary/secondary, or border). This means that if you have a site with a white background and a green foreground, the `-background-light` color variable becomes unnecessary since its impossible to have a color lighter than white. Instead, only foreground colors should come with a `-dim` suffix. 

> Are we comfortable shipping this without warning the user when they create an accessibility issue, say through a low-contrast background : primary color combination?

Yeah, I’m personally okay with this. Once a user starts to customize their own colors, I’m not sure themes should still be responsible for guaranteeing a11y. Thats just my personal view, but an a11y warning script feels like a semi-happy medium.

> Rather than (or in addition to) adding these features to the customizer, we could integrate some color processing utilities with the child theme build tools also.

Hmmm, thats not a bad idea. Definitely something for later down the line, but I could see that saving even more dev time for sure. 👍

allancole:
So I stumbled on another color contrast PHP script that works a bit better than the script I was using before. It’s definitely an improvement contrast-wise and should feel much closer to the defaults although it also doesn’t guarantee a11y-friendly colors. Check it out and let me know what you think, @kjellr @jffng 

> Also: Is there any way we can keep the original colors unless a user has manually changed a color from its default? So for example, if someone were to choose the "Custom" radio button, but then not touch any of the color controls, we'd still use the defaults for each. And if the user were to only adjust the Primary Color, then we'd only generate the --global--color-primary-hover value, and leave the rest as is.

Added a fix for this as well.

kjellr:
This is looking pretty good, @allancole! I made a _slight_ adjustment to the opacity of the generated colors, and I think they look good now. 👍 

> Also: Is there any way we can keep the original colors unless a user has manually changed a color from its default? 

I realized a weird part of this: If you only update the secondary color, the hover state for all primary color items looks really weird. Any idea how to get that working better? 

light-hover

> This may need further discussion but I think because of how that light-tan color is being using in this theme it might be better to add a new --global--color-tertiary variable, rather than tying it to the background-color variant variables...
> This is partly because the -light suffix is misleading. It should probably be renamed to -dim ...

Yes, I think that's a good idea. Do you think that should be handled here or in another PR? Maybe the tertiary color should be added here, since it will also become editable here? Or else we could just incorporate the current `--global--color-background-light` value and make that editable here again, and change all this separately?

In any case, I see that for now, the color completely disappears currently when you activate the custom palette. So we should fix that at least. 😅

tertiary-color

> Are we comfortable shipping this without warning the user when they create an accessibility issue, say through a low-contrast background : primary color combination?

To loop back to this note from @jffng: No, I'm not _super_ comfortable with it. But I do fully expect us to build this warning in, and I think we should be able to do so shortly after we package up this 1.0 release. Jeff, would you mind doing a little bit of digging to gauge the feasibility of that? Does Gutenberg use a resource we can lean on to make this happen?

varya/classes/class-varya-custom-colors.php
```diff
@@ -0,0 +1,310 @@
...
+	/**
+	 * Auto-set color luminance for hover and selection colors
+	 *
+	 * Lightens/darkens a given colour (hex format), returning the altered colour in hex format.7
+	 *
+	 * @param str $fg_hex Colour as hexadecimal (with or without hash);
+	 * @percent float $percent Decimal ( 0.2 = lighten by 20%(), -0.4 = darken by 40%() )
+	 * @return str Lightened/Darkend colour as hexadecimal (with hash);
+	 */
+	function varya_color_midpoint( $fg_hex, $bg_hex ) {
```
kjellr, line 57: I _think_ this function can be removed since we now use `varya_color_blend_by_opacity()`, right?

kjellr:
Also — I just tried this on a fresh site and I'm seeing this error: 

Screen Shot 2020-04-23 at 7 26 02 PM

allancole:
Ahh, interesting that invalid hex color issue should be a quick fix. I’ll sort it out.

allancole:
> Yes, I think that's a good idea. Do you think that should be handled here or in another PR? Maybe the tertiary color should be added here, since it will also become editable here? Or else we could just incorporate the current --global--color-background-light value and make that editable here again, and change all this separately?

I think we might have to go with the latter. Once we add this new variable, we’ll need to update quite a few Sass files, so lets do it in a separate PR.

jffng:
This is working well in my testing! 

One thing that you guys probably noticed / discussed is that when you use a text color close to white, the generated hover color doesn't come through:

Kapture 2020-04-24 at 16 19 50

-----

> If you only update the secondary color, the hover state for all primary color items looks really weird.

Every time the secondary color changes, could you run `varya_color_blend_by_opacity` on the primary color to generate new hover and selection colors ?

In general, I'm not advocating that every edge case / color scheme needs to be solved for.

-----

Re: accessibility warnings:

> would you mind doing a little bit of digging to gauge the feasibility of that? Does Gutenberg use a resource we can lean on to make this happen?

Sure thing, I'll open an issue and do some research.

allancole:
> If you only update the secondary color, the hover state for all primary color items looks really weird.
>> Every time the secondary color changes, could you run `varya_color_blend_by_opacity()` on the primary color to generate new hover and selection colors ?

@jffng this is gonna be a tricky one :-). The reason this is happening is because the theme can set every single color explicitly, while the customizer options have to work programmatically so users only have to set 5 or 6 colors instead of all 18. 

The root cause of this is the “Botanical” style borrows the `secondary` color value and uses it as the `primary-hover` color which is perfectly fine but not every child theme will work this way. 

I think the best way forward is to make it so that when you choose to use the Custom Color radio button, the “Botanical”-specific hover colors, get replaced by the auto-generated hover colors, immediately. I’ll need to rework the PHP a bit to make this happen, but anything else would distort how the system works in other child themes that set default colors differently than the Botanical default.

jffng:
After more time debugging than I'd like to admit, I pushed a fix to suppress the warnings I was getting due to invalid / null `$foreground` color values. 

I think this should be ready to go.

PR #115, opened 2020-04-22
Update Varya Elevation

kjellr:
The elevation variables in Varya are supposed to progress from least shadow to most shadow. Currently, they do not: 

Screen Shot 2020-04-22 at 2 20 20 PM

This PR adjust those values so there is a more natural progression: 

Screen Shot 2020-04-22 at 2 20 34 PM

---

It also adjusts the elevation level of the main navigation so that the shadow is a little more refined. 

Before: 

Screen Shot 2020-04-22 at 2 19 28 PM

After: 

Screen Shot 2020-04-22 at 2 17 29 PM

### DIFF
--- a/varya/assets/sass/child-theme/variables.css
+++ b/varya/assets/sass/child-theme/variables.css
@@ -68,10 +68,10 @@
 	--global--spacing-horizontal: 25px;
 	--global--spacing-vertical: 30px;
 	/* Elevation */
-	--global--elevation-1: 0px 0px 4px 2px rgba( 0, 0, 0, 0.2 );
-	--global--elevation-2: 0px 0px 8px 2px rgba( 0, 0, 0, 0.2 );
-	--global--elevation-3: 2px 2px 4px 2px rgba( 0, 0, 0, 0.2 );
-	--global--elevation-4: 2px 2px 8px 0px rgba( 0, 0, 0, 0.2 );
+	--global--elevation-1: 1px 1px 3px 0px rgba( 0, 0, 0, 0.2 );
+	--global--elevation-2: 1px 1px 4px 0px rgba( 0, 0, 0, 0.2 );
+	--global--elevation-3: 2px 2px 6px 0px rgba( 0, 0, 0, 0.2 );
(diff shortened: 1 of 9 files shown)


kjellr:
I guess one other question is... do we need all these variables? The only place that uses elevation is this menu dropdown.

allancole:
> do we need all these variables? The only place that uses elevation is this menu dropdown.

While Varya itself only uses elevation in the menu, I imagine child themes could potentially use these for more than that. Buttons, cover blocks styles and a bunch of mobile UI stuff (a la material design) come to mind. I’d like to keep them because I know I’d use them but I do hear your point in that these are kind of just there and not really integrated to the system. So I guess maybe they can be removed.

kjellr:
I think if a theme wanted to use more levels of elevation, they could write their own in. I'll remove these for now.

kjellr:
I've removed all the elevations and replaced with a single, new `--primary-nav--box-shadow` variable. Would one of you mind giving it one last look?

varya/assets/sass/child-theme/variables.css
```diff
@@ -67,13 +67,6 @@
 	--global--spacing-measure: unset;
 	--global--spacing-horizontal: 25px;
 	--global--spacing-vertical: 30px;
-	/* Elevation */
-	--global--elevation-1: 0px 0px 4px 2px rgba( 0, 0, 0, 0.2 );
```
jffng, line 71: The nav block will also have a hover state, and it would be nice to leverage the same elevation variable. Unless we don't plan on adding any elevation there, but that seems inconsistent.

kjellr: Hmm, you're right. Maybe it makes sense to have just a single box-shadow rule?

kjellr: In any case, maybe we should save this for when we style that in a PR.

jffng: Yeah I think saving one `--global--elevation` variable would be useful!

kjellr:
Alright, @jffng — I've removed all of the elevation rules except for one, and this should be ready for one last review. 👍

PR #85, opened 2020-04-14
Varya: General system audit

allancole:
This PR audits the entire Varya system as a whole to optimize it for TRT requirements and general code cleanup. 

- Removes various unnecessary filters and functions leftover from Twenty Nineteen and `_s`.
- Unify all instances of `main-menu` and `main-navigation` to use `primary-` prefix for consistency across name spaces.
- Add download icon to media attachment page (requires #52)
- Remove custom social-menu wrapper class.
- Dequeue Gutenberg theme styles in favor of our own.
- Reset border style on `hr` and separator block
- Cleans up responsive rules to better support block alignments across different blocks.

PR #123, opened 2020-04-23
Fix margins on left and right aligned blocks 

### LINKED ISSUE #104: Varya: Left and right alignments are missing margins in the editor
Editor: 

Screen Shot 2020-04-16 at 6 35 48 PM
Screen Shot 2020-04-16 at 6 35 38 PM

Front end: 

Screen Shot 2020-04-16 at 6 37 00 PM

Screen Shot 2020-04-16 at 6 37 11 PM

jffng:
Fixes #104.

**Before**
Screen Shot 2020-04-23 at 6 06 40 PM

**After**
Screen Shot 2020-04-23 at 6 08 47 PM

The issue of alignments is tricky because they are not handled consistently. Sometimes Gutenberg will add `data-align="left"` or a `.alignleft` class, it just depends on the block.

### DIFF
--- a/varya/assets/sass/base/_editor.scss
+++ b/varya/assets/sass/base/_editor.scss
@@ -50,16 +50,6 @@ a {
 		font-family: var(--global--font-secondary);
 		font-size: var(--global--font-size-md);
 	}
-
-	&.alignleft {
-		margin: 0;
-		margin-right: var(--global--spacing-horizontal);
-	}
-
-	&.alignright {
-		margin: 0;
-		margin-left: var(--global--spacing-horizontal);
-	}
 }
 
 div[data-type="core/button"] {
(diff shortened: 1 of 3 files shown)


varya/assets/sass/base/_editor.scss
```diff
@@ -50,6 +50,16 @@ a {
 		font-family: var(--global--font-secondary);
 		font-size: var(--global--font-size-md);
 	}
+
```
allancole, line 53: The `.alignleft` and `.alignright` utility classes are all located in `/sass/blocks/utilities/_style.scss`. If this isn’t fixed in #85, can you move this improvement there?

jffng: I checked out #85 and the issue was still there.

I'll move these styles over to `/sass/blocks/utilities/_style.scss` 👍

Issue #84, opened 2020-04-14
Varya: WooCommerce duplicate mini-cart issue in Customizer.


allancole:
When making _any_ change to the menu using the customizer, a duplicate mini-cart gets repeatedly rendered into the dom after every change.

Screenshot:

![2020-04-14 12 16 52](https://user-images.githubusercontent.com/709581/79248461-e18c4480-7e49-11ea-88cd-509223d69dc1.gif)

allancole:
One way to fix this could be to remove the mini-cart entirely from all customizer views. I started a PR for [that here](https://github.com/Automattic/themes-workspace/tree/fix/issue-84), but halfway through, I realized it may be too aggressive. WooCommerce has its own Customizer panel and its possible to add/remove items from the cart right in the customizer and it updates the  mini-cart as expected. 

May need to look into another approach that captures the Menu’s [selective refresh](https://make.wordpress.org/core/2016/02/16/selective-refresh-in-the-customizer/) call and prevents the mini-cart from getting duplicated. 

Another way we could do it is to use Javascript to generate the WC menu instead of PHP, which would make it easier to tap into the selective refresh call, but also require us to load another file on every page. 

@jeffikus maybe you have some input/thoughts on this?

PR #124, opened 2020-04-23
Varya: Fix duplicate mini-cart issue in customizer

allancole:
This PR prevents the mini-cart from getting duplicated when the menu gets refreshed in the customizer.

Before|After
------------|------------
varya-menu-refresh-before|varya-menu-refresh-after

To test, make sure Woo is activated, visit the customizer and make any change to the menu.

Fixes: #84

### DIFF
--- a/varya/assets/js/customize-preview-woocommerce.js
+++ b/varya/assets/js/customize-preview-woocommerce.js
@@ -8,6 +8,7 @@ jQuery( function( $ ) {
 
 	// Remove mini-cart
 	function resetWooMiniCart( container ) {
+
 		var wooMenuContainer = container.siblings( ".woocommerce-menu-container[title|='Shift']" );
 		var wooMenuButton    = container.siblings( "#toggle-cart[title|='Shift']" );
 
@@ -15,6 +16,7 @@ jQuery( function( $ ) {
 		wooMenuButton.remove();
 	}
 
+	// Listen for preview menu refresh
 	$( document ).on( 'customize-preview-menu-refreshed', function( e, params ) {
 		if ( 'primary' === params.wpNavMenuArgs.theme_location ) {
 			resetWooMiniCart( params.newContainer );
--- a/varya/assets/js/customize-preview-woocommerce.js
+++ b/varya/assets/js/customize-preview-woocommerce.js
@@ -7,7 +7,7 @@
 jQuery( function( $ ) {
 
 	// Remove mini-cart
-	function resetWooMenu( container ) {
+	function resetWooMiniCart( container ) {
 		var wooMenuContainer = container.siblings( ".woocommerce-menu-container[title|='Shift']" );
 		var wooMenuButton    = container.siblings( "#toggle-cart[title|='Shift']" );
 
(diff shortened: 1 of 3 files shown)


varya/assets/js/customize-preview-woocommerce.js
```diff
@@ -0,0 +1,25 @@
...
+
+		var wooMenuContainer = container.siblings( ".woocommerce-menu-container[title|='Shift']" );
+		var wooMenuButton    = container.siblings( "#toggle-cart[title|='Shift']" );
+
+		wooMenuContainer.remove();
+		wooMenuButton.remove();
+	}
+
+	// Listen for preview menu refresh
+	$( document ).on( 'customize-preview-menu-refreshed', function( e, params ) {
```
jffng, line 20: This function removes the Woo mini cart every time the customizer preview refreshes. So where is the logic that adds the woo mini cart on every refresh? It seems like it would live in the Woo plugin itself. 

If that's a Woo bug (seems like it is), we could need to remove this code if they fix the bug. Otherwise the mini-cart would disappear by this function. Am I understanding the issue correctly?

allancole: Ahhh right, so this file `customize-preview-woocommerce.js` is specifically for dealing with WooCommerce in the customizer, but the mini-cart is actually injected in this file: 

https://github.com/Automattic/themes-workspace/blob/f287bf39693ff136b3774cbd4012c22b44b6ec99/varya/inc/woocommerce.php#L216 

It runs on both the customizer and the frontend. The reason the cart gets duplicated is because this function is used in a filter on `add_filter( 'wp_nav_menu_items', 'varya_add_cart_menu', 10, 2 );`. Every time the menu gets changed in the customizer, selective-refresh has to re-run `'wp_nav_menu_items'` so a second menu appears. 

Another way to solve this could be to rewrite the mini-cart so it isn’t appended directly on the main menu. Then it would totally ignores the selective refresh calls. Could be a pretty big refactor but maybe it’s worth it?

jffng: I see...

> Another way to solve this could be to rewrite the mini-cart so it isn’t appended directly on the main menu.

Where / how would we add it instead? 

Alternatively, could we add a condition in `varya_add_cart_menu` to check if the nav already contains an instance of `woocommerce-menu-container` or something unique?

allancole: > Where / how would we add it instead?

So instead of appending the mini-cart to the primary menu as the last list-item, we could make it it’s own, separate `<nav>` menu and completely disconnect it from the primary menu so its easier to deal with. Might require more CSS, but thinking about it again, it’s probably more preferable to have the mini-cart separate so child themes can do more with it. 

Curious what you think @kjellr?

kjellr: It probably makes sense to have it separate, but I don't have a strong preference.

allancole: Yeah, after thinking on this a bit more, I definitely think we should make the WooCommerce menu completely separate—its just more flexible and less opinionated that way. We can also keep the markup similar so we don’t have to redo a ton of CSS styles.

kjellr: It looks like aside from reaming the theme, this is the only bug that's holding us up from a 1.0 release. Should this PR land as is and then we re-assess the completely separate WooCommerce menu in our next batch of updates?

kjellr:
Closing in favor of #159

PR #126, opened 2020-04-24
Clean version of #85

jffng:
Clean PR to more easily view changes from #85 by @allancole:

- Removes various unnecessary filters and functions leftover from Twenty Nineteen and _s.
- Unify all instances of main-menu and main-navigation to use primary- prefix for consistency across name spaces.
- Add download icon to media attachment page (requires #52)
- Remove custom social-menu wrapper class.
- Dequeue Gutenberg theme styles in favor of our own.
- Reset border style on hr and separator block
- Cleans up responsive rules to better support block alignments across different blocks.


varya/assets/sass/blocks/file/_style.scss
```diff
@@ -17,12 +17,10 @@
 	}
 
 	.entry-content &.alignleft {
-		@extend %responsive-alignright;
-		margin-right: var(--global--spacing-horizontal);
+		@extend %responsive-alignleft;
```
jffng, line 20: Why is it that this block stills need the alignment rules here? Shouldn't the alignment utility classes take care of it? Same question for the pullquote and video blocks.

varya/assets/sass/structure/_responsive-logic.scss
```diff
@@ -181,9 +181,9 @@ $breakpoint_xxl: 1024px;
 @include media(mobile) {
 	%responsive-alignleft {
 		/*rtl:ignore*/
-		margin-left: auto;
+		margin-left: var(--responsive--alignleft-margin);
```
jffng, line 184: I'm encountering a small bug when I use the `alignleft` class:

Left aligned cover block | No alignment cover block
-------- | --------
Screen Shot 2020-04-24 at 12 08 33 PM|Screen Shot 2020-04-24 at 12 08 54 PM


The left aligned cover block is slightly out of alignment with the default content width. Should this be a separate issue that tackles the way the `--responsive--alignleft-margin` value is calculated?

varya/assets/sass/blocks/cover/_style.scss
```diff
@@ -72,29 +72,6 @@
 		}
 	}
 
-	/**
-	 * Block Options
-	 */
-	.entry-content > &.alignright {
```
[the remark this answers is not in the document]
allancole: Apologies on the late reply but yes, that is correct. With the latest markup improvements in Gutenberg, it means we don’t need to override utility classes with block-specific styles as much.

varya/assets/sass/blocks/file/_style.scss
allancole: This might be able to be removed too. Need to test it with Gutenberg 8.0 to be sure, but at the time I added this, it was still necessary.

varya/assets/sass/structure/_responsive-logic.scss
allancole: Hmmm. Having trouble replicating this one. Could double check that this is still happening for you with the latest version of Gutenberg?

varya/assets/sass/blocks/file/_style.scss
jffng: Tested, and these are still necessary.

varya/assets/sass/structure/_responsive-logic.scss
jffng: This is still happening, but it's not a result of this PR.

Issue #116, opened 2020-04-22
Varya: update navigation block styles


jffng:
Some updates are needed to the navigation block styles. The first two should be fixed, the third is optional.

1. The navigation block styles need to match in the front-end and editor.

Editor | Front-end
----- | -----
<img width="563" alt="Screen Shot 2020-04-22 at 2 59 37 PM" src="https://user-images.githubusercontent.com/5375500/80022403-02881180-84aa-11ea-9064-61ea0c8afa20.png">|<img width="616" alt="Screen Shot 2020-04-22 at 2 51 45 PM" src="https://user-images.githubusercontent.com/5375500/80021645-cef8b780-84a8-11ea-8e32-6e566bf7aafd.png">

-------------

2. Submenus have an overlay / transparency issue (both editor and front-end)
<img width="646" alt="Screen Shot 2020-04-22 at 3 09 25 PM" src="https://user-images.githubusercontent.com/5375500/80023277-47f90e80-84ab-11ea-8b1d-5a282b8fea93.png">

-------------

3. We could consider updating the nav block styles to match the primary nav. 

Primary nav | Nav block
----- | -----
<img width="398" alt="Screen Shot 2020-04-22 at 3 06 21 PM" src="https://user-images.githubusercontent.com/5375500/80023099-04060980-84ab-11ea-8c7a-69f87ae8e7cb.png">|<img width="535" alt="Screen Shot 2020-04-22 at 3 06 17 PM" src="https://user-images.githubusercontent.com/5375500/80023118-0cf6db00-84ab-11ea-8f8e-8b5b6d5d81e4.png">

PR #130, opened 2020-04-27
Varya: Add nav block styles

jffng:
Addresses #116.

This PR adds styles to navigation block to match what we have for the primary navigation.

**Front-end**

Before | After
------- | -------
Screen Shot 2020-04-27 at 6 37 48 PM|Screen Shot 2020-04-27 at 6 37 30 PM

**Editor**
Before | After
------- | -------
Screen Shot 2020-04-27 at 6 39 00 PM|Screen Shot 2020-04-27 at 6 37 09 PM

A few considerations:
- It only adds a handful of editor styles as the UI for adding nav items is complex.
- I did not replace the icons around nested menus because they are inline SVG elements generated by the block.
- I did not do anything to specifically address the mobile use case, because supporting the ability to add several fullscreen mobile menus via the nav block would require some additional front-end logic. But we may want to revisit this assumption.

### DIFF
--- a/varya/assets/sass/variables-editor.scss
+++ b/varya/assets/sass/variables-editor.scss
@@ -25,6 +25,8 @@
 	@include quote-variables();
 	@include separator-variables();
 	@include utilities-variables();
+	/* Components */
+	@include header-variables();
 	/* Vendors */
 	@include woocommerce-variables();
 }
--- a/varya/assets/css/style-editor.css
+++ b/varya/assets/css/style-editor.css
@@ -794,6 +794,26 @@ dt {
 	color: currentColor;
 }
 
+.wp-block-navigation .wp-block-navigation__container {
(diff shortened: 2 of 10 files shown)


varya/assets/sass/variables-editor.scss
```diff
@@ -25,6 +25,8 @@
 	@include quote-variables();
 	@include separator-variables();
 	@include utilities-variables();
+	/* Components */
+	@include header-variables();
```
jffng, line 29: I'm importing the header variables here to get access to the `--primary-nav` variables in the editor. Might be overkill since it also includes a lot of branding variables?

kjellr: I think it's okay, since we also may want to use some of these variables later on in the Site Title + Site Description blocks, etc.

Issue #92, opened 2020-04-16
Varya: Figure out a way to include a narrow color stripe behind group block content


kjellr:
As per the comps here: 

![Screen Shot 2020-04-16 at 3 32 48 PM](https://user-images.githubusercontent.com/1202812/79498565-88621380-7ff7-11ea-8c79-0a4ec9ee6a04.png)

Could be a group block style with some weird background position adjustments? Or perhaps a gradient background with 3 color stops (white/pink/white)? Needs some experimentation.

jffng:
> Could be a group block style with some weird background position adjustments?

I tried this out in [an earlier PR](https://github.com/Automattic/themes-workspace/pull/16/commits/ad346582e058084f4995ecf98e102ab5249c0e6b#diff-bebd17ac9d461436bb9e845a929402f9R106-R112
) as a group block style. The effect can somewhat be achieved with negative margins on the container and inner elements.

A gradient treatment might be more adaptive though, so you're not relying on percentages or v-units.

PR #131, opened 2020-04-28
Add stripe gradient preset (#92)

jffng:
Addresses #92, adding a stripe background as a gradient preset. 

We may want to finesse the height of the strip.

Screen Shot 2020-04-28 at 11 51 19 AM

Noting that it still encounters this same Gutenberg bug: https://github.com/WordPress/gutenberg/issues/21831.

### DIFF
--- a/varya/functions.php
+++ b/varya/functions.php
@@ -203,27 +203,32 @@ if ( ! function_exists( 'varya_setup' ) ) :
 				array(
 					'name'     => __( 'Half cream, half brick', 'en' ),
 					'gradient' => 'linear-gradient(to top, #A36265 50%, #FAFBF6 50%)',
-					'slug'     => 'hard-horizontal-inverted'
+					'slug'     => 'hard-horizontal-inverted',
 				),
 				array(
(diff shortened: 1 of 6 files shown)


varya/functions.php
```diff
@@ -203,27 +203,32 @@ function varya_setup() {
 				array(
 					'name'     => __( 'Half cream, half brick', 'en' ),
 					'gradient' => 'linear-gradient(to top, #A36265 50%, #FAFBF6 50%)',
-					'slug'     => 'hard-horizontal-inverted'
+					'slug'     => 'hard-horizontal-inverted',
```
jffng, line 206: My php linter was yelling at me to add these commas.

jffng:
I just realized that none of the gradients use the theme's color variables. I think they should, even though the gradient preview would still require manual updating via `functions.php`. 

Any thoughts or objections?

kjellr:
Yeah, they should —  we should mimic the same thing we do for the custom colors.

jffng:
Ok, I'll merge this and open a separate PR for handling variables + custom colors.

PR #132, opened 2020-04-28
Varya: Use variables in gradient presets

jffng:
This PR does the following:

- Replaces gradient preset color values with their respective variables.
- Renames the presets to be color agnostic.
- Checks the `custom_colors_active` theme mod to see if custom colors have been set. (Depends on #113)

This should have no impact on the theme visually.


jffng:
Tested this after having set custom colors, and it works like a charm:

Screen Shot 2020-05-08 at 11 14 27 AM

PR #135, opened 2020-05-01
Varya: Set max width of author bio

### LINKED ISSUE #134: Varya: author bio max width
The width of the author bio should be set the theme's default width, but currently it is full width:

Screen Shot 2020-04-29 at 4 26 30 PM

jffng:
Fixes #134. While we're here, are the "Published by author name" font styles acceptable?

Before
Screen Shot 2020-05-01 at 12 05 35 PM

After
Screen Shot 2020-05-01 at 12 03 41 PM

### DIFF
--- a/varya/style-rtl.css
+++ b/varya/style-rtl.css
@@ -3394,6 +3394,10 @@ nav a {
  * Author
  */
 /* Author description */
+.author-bio {
+	max-width: var(--responsive--aligndefault-width);
+}
+
 .site-main > article > .author-bio {
 	margin-top: calc(2 * var(--global--spacing-vertical));
 }
--- a/varya/style-rtl.css
+++ b/varya/style-rtl.css
@@ -3395,6 +3395,7 @@ nav a {
  */
 /* Author description */
 .author-bio {
+	font-size: var(--global--font-size-xs);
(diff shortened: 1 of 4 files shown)


varya/style-rtl.css
```diff
@@ -2534,32 +2534,32 @@ table th,
 	color: var(--global--color-primary);
 }
 
-.is-tiny-text,
-.has-tiny-font-size {
+:root .is-tiny-text,
```
kjellr, line 2537: I think this file isn't properly compiled.

jffng:
I'll go ahead and work on it here.

jffng:
Ready for another review 👍 @kjellr 

Screen Shot 2020-05-07 at 2 54 42 PM

varya/template-parts/post/author-bio.php
```diff
@@ -9,19 +9,21 @@
 
 if ( (bool) get_the_author_meta( 'description' ) ) : ?>
 <div class="author-bio">
+	<?php
+	printf(__( 'Published by', 'varya' ));
```
kjellr, line 13: Does this need to use `printf()` since it doesn't actually need a variable anymore? I think it could just use `_e()` as seen in the "View more posts" link down below.

Issue #97, opened 2020-04-16
Rename theme


kjellr:
We should rename Varya to “Root” to play off the plant-based theme of these mockups, to speak to the fact that this is the “root” of our team's themes moving forward, and also to reference the fact that CSS variables are assigned to the `:root`.

TBD: Whether or not the name is in conflict with anything already out there ([one example here](https://roots.io/sage/)).

kjellr:
Timeline-wise, let's tackle this close to release. 👍

kjellr:
In case this name doesn't work for some reason, here is a dump of some other options: 

- Seedling
- Stem
- Leaf
- Leaves
- Bud
- Flower
- Pollen

(I kind of like Seedling, actually)

jffng:
In order of my preferences:

1. Root
2. Stem
3. Seedling

kjellr:
Following another round of searching for potential name conflicts, the leading candidate is now:

🌱 Seedlet

That's not a common variation of the word "seed", which is a plus for us. According to [the dictionary](https://www.merriam-webster.com/dictionary/seedlet), it means "a small seed", which is apt and also cute.

kjellr:
@jffng Feel free to start a PR to change it to Seedlet when you have a moment.

jffng:
Yep I will do this!

PR #137, opened 2020-05-01
Varya: Rename theme to Root

jffng:
Addresses #97. 

Should have no effect on theme behavior, but you will need to reactivate it on any test sites.


kjellr:
Closing this one, in favor of #166

Issue #129, opened 2020-04-24
Varya: customizer color contrast accessibility


jffng:
Follows #94 and #113. 

It would be ideal to warn the user when they create inaccessible color combinations using the customizer.

In Gutenberg, these warnings appear automatically in the block's color settings:

<img width="282" alt="Screen Shot 2020-04-24 at 4 30 38 PM" src="https://user-images.githubusercontent.com/5375500/80254389-1027cd80-8649-11ea-9cf7-3cf9c55b435b.png">

To-do:
- [x] Research and identify color processing utilities that can be leveraged in the customizer
- [ ] Determine which combinations of variables need to be evaluated
- [ ] Implement the color processing utilities and subsequent warnings

kjellr:
https://github.com/soderlind/2016-customizer-demo
This appears pretty close to what we're looking to do, and it's GPL-licensed:

![wcag-color-contrast](https://user-images.githubusercontent.com/1202812/81074418-f8591080-8eb6-11ea-9976-26ff177cfd6f.gif)

jffng:
Working off of #113, I've got the bones of this working, but the UX is pretty miserable:

![Kapture 2020-05-07 at 12 15 57](https://user-images.githubusercontent.com/5375500/81318820-cb466280-905c-11ea-9826-1d6214212b31.gif)

A few thoughts:
- I wish the notification weren't so disruptive, and that I could control its position. Maybe I can with some customizer specific CSS?
- How should we handle displaying multiple conflicts? I feel like there should only be a single notification, noting all of the violations.
- What should this text say?

These feel like questions to address in a PR, but would like to get #113 merged first.

kjellr:
> How should we handle displaying multiple conflicts? I feel like there should only be a single notification, noting all of the violations.

Yeah, just one notification is fine. More than that is overkill. 

> What should this text say?

I think we should just use the same language that's in the Gutenberg one (or something really close to it): "This color combination may be hard for people to read. Try using a brighter background color and/or a darker text color."

PR #138, opened 2020-05-07
Varya: add custom color contrast warnings

jffng:
Follows #113, addresses #129. 

This PR adds contrast warnings to the customizer. 

Kapture 2020-05-08 at 9 12 00

Currently it compares the following color combinations and issues a warning if AA standards are not met:
- Primary -> Background, Background Light
- Foreground -> Background, Background Light
- Background -> Primary, Foreground
- Background Light -> Primary, Foreground


jffng:
It bothers me that the notification appears where it does, shifting the position of the color picker UI.

But I'm not sure how to change this. Is there a way to load CSS from the theme to change how the customizer appears?

varya/assets/js/customizer-validate-wcag-color-contrast.js
```diff
@@ -0,0 +1,214 @@
+/**
+ * @author Per Soderlind
```
kjellr, line 2: Would you mind adding license info for this to the [`readme.txt`](https://github.com/Automattic/themes-workspace/blob/b9fcb3331d76efd942b9e553ca91557bf3523494/varya/readme.txt#L44) and the [`style.css`](https://github.com/Automattic/themes-workspace/blob/b9fcb3331d76efd942b9e553ca91557bf3523494/varya/style.css#L36) header?

varya/assets/js/customizer-validate-wcag-color-contrast.js
```diff
@@ -0,0 +1,214 @@
...
+				var color_to_compare = $( '#customize-control-' + color_to_compare_id + ' .wp-color-result')[0].style.backgroundColor.match(/\d+/g);
+				var contrast	= self.rgb( [ parseInt(color_to_compare[0]), parseInt(color_to_compare[1]), parseInt(color_to_compare[2]) ], self.hexRgb( value ) );
+				var score		= self.score( contrast );
+				if ( score === 'Fail' ){
+					failsWCAG = true;
+				} 
+			});
+
+			if ( failsWCAG ){
+				var validationWarning = new api.Notification( code, { message: 'This color combination may be hard for people to read. Try using a brighter background color and/or a darker text color.', type: 'warning' } );
```
allancole, line 50: We may need to do some fancy PHP/JS translation escaping here so that this text can be translated more easily. Its kind of a annoying to do, but I’m sure it’s something the TRT team would catch. I’ll send you some links that show how we do it.

Also wondering if we should change `Try using a brighter background color and/or a darker text color.` to this: `Try using a brighter background color and/or a darker **foreground** color.` since the text color is called foreground in the UI.

varya/classes/class-varya-custom-colors.php
```diff
@@ -262,6 +267,31 @@ function varya_editor_custom_color_variables() {
...
+		$deps 	= [ 'customize-controls' ];
+
+		$exports = [
+			'validate_color_contrast' => [
+				// key = current color control , values = array with color controls to check color contrast against
+				'varya_--global--color-primary' => [ "varya_--global--color-background", "varya_--global--color-background-light" ],
+				'varya_--global--color-foreground' => [ "varya_--global--color-background", "varya_--global--color-background-light" ],
+				'varya_--global--color-background' => [ "varya_--global--color-primary", "varya_--global--color-foreground" ],
+				'varya_--global--color-background-light' => [ "varya_--global--color-primary", "varya_--global--color-foreground" ],
+			],
```
allancole, line 288: Also noting that we recently realized that `varya_--global--color-background-light` isn’t very useful conceptually. It will likely become `varya_--global--color-tertiary` or `varya_--global--color-accent` so that theme designers can choose to use an extra color in child themes similar to a primary or secondary color. 

This means the colors won’t need to be evaluated against `varya_--global--color-background-light` at all. Instead, I’m thinking we’ll only need to evaluate the colors against the `varya_--global--color-background` color. We can do this in a separate PR after this is merged.

kjellr:
Two things to note: 

It would be great if we could have just one of these notices appear, and only appear at the very bottom. 

Also, I'm having trouble getting the notices to go away after I've corrected the problem: 

Screen Shot 2020-05-08 at 1 53 29 PM

jffng:
I made the following changes to this PR:

- Localization of the warning text
- A single warning notification placed near the top when a violation occurs
- The warning is dismissed when the violation is resolved

Screen Shot 2020-05-11 at 3 34 18 PM

varya/classes/class-varya-custom-colors.php
```diff
@@ -262,6 +267,34 @@ function varya_editor_custom_color_variables() {
...
```
allancole, line 288: Looking at this again, I think we can simply compare all of the colors against the background-color (accept the background itself which can be compared with the foreground). Otherwise, when we compare the colors with more than just one color, it gets tricky to figure out which colors is actually causing the warning. If we change it to something like this, less warning pop up and when they do, it clearer which color option is causing it.

```suggestion
			'validate_color_contrast' => [
				// key = current color control , values = array with color controls to check color contrast against
				'varya_--global--color-primary'          => [ "varya_--global--color-background" ],
				'varya_--global--color-secondary'        => [ "varya_--global--color-background" ],
				'varya_--global--color-background-light' => [ "varya_--global--color-background" ],
				'varya_--global--color-foreground'       => [ "varya_--global--color-background" ],
				'varya_--global--color-background'       => [ "varya_--global--color-foreground" ],
			],
```

Also, we’re considering turning `varya_--global--color-background-light` into `varya_--global--color-accent` or `varya_--global--color-tertiary`. It’ll basically be a dormant utility class that child themes can tap into to add a third color if a design calls for it. The `background-light` and `background-dark` colors actually don’t make sense in the system.

kjellr: Agreed, let's only compare to the background. 👍 

Until we sort out a definite name/direction for the `background-light` option, I think we should just leave contrast tests off of that one. Otherwise it just gets too complicated. We can add that in later.

kjellr:
I just gave this a try, and it feels _really_ close. Let's just make that change Allan suggested above, and then I think this should be good to merge. 🎉

jffng:
Thank you both for the reviews!

PR #141, opened 2020-05-09
sk8prk: initial structure

jffng:
This PR creates the scaffolding for a Varya child theme, sk8prk designed by @melchoyce. 

Screen Shot 2020-05-13 at 6 21 03 PM

I've started off with a structure that mirrors the advanced child theme that @allancole had started for Nwaneri, with a handful of modifications. 

**Child Theme Structure**

```
sk8prk
|__ style.css — stylesheet used by the front-end and editor
|__ variables.css — varya variable overrides used by the front-end and editor
|__ functions.php — load fonts, set editor font sizes & palette 
|__ header.php — handful of changes needed for the header
|__ assets
    |__ sass — structural styles
    |__ block-extends — block styles + scripts for registering style variations
|__ template-parts — handful of changes needed mainly for layout
```

I'd love to know how the general structure of the child theme sits with you. Also going to leave some comments in line where I made some assumptions.

What I think is pretty cool is that by adding a stylesheet of about 300 lines plus overriding about 70 variables, you can get this far: https://sk8prk.mystagingwebsite.com 

There's still a lot to do, but I figured this is a good point to get feedback on the structure before moving too far ahead.


.github/workflows/deploy-sk8prk.yml
```diff
@@ -0,0 +1,23 @@
+name: Deploy sk8prk
```
jffng, line 1: This action deploys the theme automatically to the pressable site.

sk8prk/assets/block-extends/_image.scss
```diff
@@ -0,0 +1,24 @@
+.wp-block-image {
+    &.is-style-duotone {
```
jffng, line 2: All of the duotone effects are created with CSS and blend-modes, and applied with block style variations.

sk8prk/functions.php
```diff
@@ -0,0 +1,160 @@
...
+function varya_sk8prk_fonts_url() : string {
+    $fonts_url = '';
+
+	$font_families   = array();
+	$font_families[] = 'family=Red+Hat+Display:ital,wght@0,900;1,900';
+	$font_families[] = 'family=Red+Hat+Text:ital,wght@0,400;0,500;0,700;1,400;1,500;1,700';
+	$font_families[] = 'display=swap';
+
+    // Make a single request for the theme fonts.
+    $fonts_url = 'https://fonts.googleapis.com/css2?' . implode( '&', $font_families );
```
jffng, line 152: Google fonts API has changed, so the request format has changed also. We can add back the `if` statements.

sk8prk/header.php
```diff
@@ -0,0 +1,77 @@
...
+	<meta name="viewport" content="width=device-width, initial-scale=1" />
+	<link rel="profile" href="https://gmpg.org/xfn/11" />
+	<?php wp_head(); ?>
+</head>
+
+<body <?php body_class(); ?>>
+<div id="page" class="site">
+	<a class="skip-link screen-reader-text" href="#content"><?php _e( 'Skip to content', 'sk8prk' ); ?></a>
+
+		<header id="masthead" class="site-header wide-max-width">
```
jffng, line 26: It seems like overkill to include this template, considering all that has changed is the class name here to `wide-max-width`.

sk8prk/functions.php
kjellr: This definitely doesn't have to happen here in this PR, but I'd love for us to try bundling the fonts to eliminate the extra call to Google.

sk8prk/assets/block-extends/_image.scss
kjellr: We should double check how these render on IE (even the later versions). I remember some issues with duotone images on Twenty Nineteen.

kjellr:
The structure here looks pretty good to me, and I'm impressed with what we can do with just a small amount of CSS. 🎉

sk8prk/functions.php
allancole: Yeah, I’ve been getting “fout” with this new fonts API. Hopefully bundling the fonts fixes it.

sk8prk/header.php
allancole: Ahhh, yeah since thats just a responsive class, you can probably add something like this to `_header.scss` and get the same result without this template file (_I think_):

```
.site-header {
    @include %responsive-alignwide-width;
}
```

allancole:
This looks super solid @jffng! Happy to see you taking the system and running with it successfully without a ton of effort :-) 

Regarding structure, I was sort of expecting something something like this for the block extend assets:

```
sk8prk
|__ assets
    |__ js  — scripts (*** extend-blocks.js would go in here ***)
    |__ sass — styles
        |__ block-extends — block extend styles
```

To me, that feels a little closer to what’s already in Varia and keeps the similar things together. I don’t have a super strong opinion on this but otherwise, I think this is a go. 👍

PR #144, opened 2020-05-11
Varya: fix custom color class bug

jffng:
This PR attempts to fix the following bug, reported by @melchoyce:

image (6)

We likely did not catch this due to differences in php versions.


kjellr:
👍 Good catch! Thanks for fixing.

Issue #133, opened 2020-04-29
Varya: add a theme screenshot


jffng:
The theme should have an appealing screenshot!

kjellr:
Fixed in #148. The new screenshot looks like this: 

![Screenshot](https://user-images.githubusercontent.com/1202812/82462228-52e29700-9a89-11ea-8365-1298f60ef89a.png)

PR #148, opened 2020-05-20
Varya: Change primary color from red to green

kjellr:
This has been simmering for a while, but before I create a screenshot for #133, I figured we'd make a final call. The current rose color has been ok, but it feels a little dreary. In keeping with the botany/plants theme, I'd love to switch to a green color. The one in here was explored previously in comps, and I think works pretty well. 

Current|Proposed
---|---
Dark Red|Teal

cc @melchoyce, @allancole, @jffng


kjellr:
Here's a broader spectrum of things that were considered. The current color is on the left, and this new green one is second from the right: 

Screen Shot 2020-05-20 at 10 05 21 AM

kjellr:
Excellent, thank you. I've updated the screenshot, so this PR will now also fix #133.

jffng:
Should the text-selection color be updated also? 

Screen Shot 2020-05-20 at 3 09 31 PM

kjellr:
Ah, right. That should be updated too. I wish we could automate that somehow. 🤔

PR #160, opened 2020-05-27
Varya: Add mobile full-width rules

kjellr:
I noticed that full-width blocks were not appearing full-width on mobile. This is because of a general padding rule around the entire page. 

This PR creates a new `%responsive-alignfull-width-mobile` extend rule, and adds that to the places where we were previously extending `%responsive-alignfull-width`. This follows the approach used for mobile rules in `%responsive-alignleft` and `%responsive-alignright`.

The new rule adds a negative margin and increases the width of the element so that it's truly full screen on small screens. 

Before|After
---|---
dotorgthemes test__p=19|dotorgthemes test__p=19 (1)

### DIFF
--- a/varya/assets/css/style-editor.css
+++ b/varya/assets/css/style-editor.css
@@ -81,9 +81,19 @@
 }
 
 .full-max-width {
-	max-width: var(--responsive--alignfull-width);
-	margin-left: auto;
+	max-width: calc(100% + (2 * var(--global--spacing-horizontal)));
+	width: calc(100% + (2 * var(--global--spacing-horizontal)));
+	margin-left: calc(-1 * var(--global--spacing-horizontal));
 	margin-right: auto;
+}
+
+@media only screen and (min-width: 482px) {
+	.full-max-width {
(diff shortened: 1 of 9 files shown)


kjellr:
Good catch. That's definitely not right. I fixed all of that in a5462df, but only by messing with a lot of selectors we're not really supposed to have to though. We should  not merge with that in here, and should instead figure out where things are broken. 

I also noticed that full-width blocks were appearing with a ton of unnecessary padding on large screens in the editor: 

Screen Shot 2020-05-28 at 7 03 11 PM

That's clearly wrong. a5462df fixes it, but again — it shouldn't need to. I'm wondering if maybe there's a Gutenberg bug that's messing with this? Varya doesn't really mess with Group block styles in the editor usually.

kjellr:
I was right — the Group block had a regression. 😄 17ab13a removes the code on our end, since it'll be fixed in Gutenberg. 

I think the Columns block one is a regression in Gutenberg as well, so I'll take that out for now.

kjellr:
Closing in favor of #170

PR #166, opened 2020-06-03
Rename Varya to Seedlet

### LINKED ISSUE #97: Rename theme

kjellr:
Closes #97

I _think_ I caught everything here. I also made changes to `varya-blocks` and `sk8prk` because they threw errors in WP-Admin if I didn't. 

It strikes me that we should've merged #160 before doing this, but I think I can get that sorted out after this one instead.


seedlet/README.md
```diff
@@ -15,93 +15,93 @@ When you reduce a theme design down to a set of systematic design decisions, you
...
-To use the system, simply duplicate the `varya-child` theme directory and rename it `my-theme-name`. You’ll also want to do a search for `varia-child` strings and replace them with `my-theme-name` as follows:
+To use the system, simply duplicate the `seedlet-child` theme directory and rename it `my-theme-name`. You’ll also want to do a search for `varia-child` strings and replace them with `my-theme-name` as follows:
 
-- Search for: `'varya-child'` and replace with: `'my-theme-name'` (with quotes).
-- Search for: `varya_child_` and replace with: `my_theme_name_`.
+- Search for: `'seedlet-child'` and replace with: `'my-theme-name'` (with quotes).
+- Search for: `seedlet_child_` and replace with: `my_theme_name_`.
 - Search for: `Text Domain: varia-child` and replace with: `Text Domain: my-theme-name` in _style.css_.
 - Search for:  `varia-child` and replace with: `my_theme_name`.
 - Search for: `varia-child-` and replace with: `my-theme-name-`.
```
allancole, line 50: A few instances of `varia-` popped up here. 
Otherwise, I think this is good to go :-)

kjellr: Oh weird! I wonder how that got through the find & replace. 🤔 

Thanks for catching!

kjellr: Should be fixed in 4b73264. Mind taking one more look?

PR #170, opened 2020-06-04
Add mobile full-width rules.

kjellr:
**⚠️ Note that this is based on the update/rename-varya-to-seedlet branch, so we'll want that merged first. ⚠️**

This is the same PR as #160, just updated to land after #166. 

---

I noticed that full-width blocks were not appearing full-width on mobile. This is because of a general padding rule around the entire page. 

This PR creates a new `%responsive-alignfull-width-mobile` extend rule, and adds that to the places where we were previously extending `%responsive-alignfull-width`. This follows the approach used for mobile rules in `%responsive-alignleft` and `%responsive-alignright`.

The new rule adds a negative margin and increases the width of the element so that it's truly full screen on small screens. 

Before|After
---|---
dotorgthemes test__p=19|dotorgthemes test__p=19 (1)


jffng:
Pushed one editor update to fix the a group blocks inner container padding to reflect the front-end:

**Before**
Screen Shot 2020-06-04 at 9 41 28 AM

**After**
Screen Shot 2020-06-04 at 9 41 38 AM

PR #172, opened 2020-06-04
Seedlet Blocks: General Improvements

kjellr:
- Defines `tagName` to implement semantic tags for the wrapper blocks. 
- Adds a nav block to the page header.
- Removes the rounded style from the main header image, since we don't do that in the main theme anymore. 
- Adds a theme screenshot + readme
- Brings the style.css header up to date.


seedlet-blocks/block-template-parts/header.html
```diff
@@ -1,9 +1,13 @@
-<!-- wp:image {"align":"center","width":128,"height":128,"sizeSlug":"large","className":"is-style-rounded"} -->
-<div class="wp-block-image is-style-rounded"><figure class="aligncenter size-large is-resized"><img src="https://cldup.com/B9dfntUFJE.png" alt="" width="128" height="128"/></figure></div>
```
kjellr, line 2: I also removed the rounded style from this image, since we removed that in the main theme.

seedlet-blocks/block-templates/singular.html
```diff
@@ -1,13 +1,13 @@
-<!-- wp:group {"align":"full","className":"site-header"} -->
+<!-- wp:group {"align":"full","className":"site-header","tagName":"header"} -->
 <div class="wp-block-group alignfull site-header"><div class="wp-block-group__inner-container"><!-- wp:template-part {"slug":"header","theme":"seedlet-blocks"} /--></div></div>
 <!-- /wp:group -->
 
-<!-- wp:group {"align":"full","className":"page-title"} -->
-<div class="wp-block-group alignfull page-title"><div class="wp-block-group__inner-container"><!-- wp:post-title /--></div></div>
+<!-- wp:group {"align":"full","className":"site-content","tagName":"main"} -->
+<div class="wp-block-group alignfull site-content"><div class="wp-block-group__inner-container">
+<!-- wp:post-title /-->
```
jffng, line 7: When wrapped in a alignfull group block, the post title shows up for me like this: 

Screen Shot 2020-06-04 at 2 39 19 PM

kjellr: Hmm weird. I pushed an update in 490dc73, but it seems like it should be unnecessary. 🤔

kjellr: Nevermind, I figured it out — Seedlet's standard wide/full styles only apply themselves to blocks inside of `.entry-content`, so they're not active here. We should probably change that in Seedlet if we can, since it'll make things way easier for full-site editing. For now, I've duplicated those styles here in the child theme.

Issue #174, opened 2020-06-04
Seedlet: Custom colors show up broken on fresh install


kjellr:
On a fresh install, all of the custom colors show up broken. This can be fixed by visiting the customizer, choosing to use non-default colors, and then reverting back to the default again. 

<img width="913" alt="Screen Shot 2020-06-04 at 10 16 06 AM" src="https://user-images.githubusercontent.com/1202812/83796677-6c7e0400-a66f-11ea-956d-0e366f7b66c0.png">

kjellr:
@jffng is digging into this, but cc-ing @allancole in case you have any immediate ideas. This issue was first noticed after we renamed the theme, but it's totally possible that it existed since we built custom colors in, and that we just never noticed.

allancole:
Ahhh, I’m about 93% sure I have a solution for this. Will try it in a PR shortly :-)

PR #175, opened 2020-06-05
Seedlet: Improve custom color conditionals.

allancole:
This PR does the following: 
- Makes sure theme mods are _not_ empty before choosing default or custom color
- Uses new conditionals in gradient color variables

To test, add and activate Seedlet on a brand new site, visit the editor, and view the colors panels of any block. You should see the full set of default colors appear instead of 6 black colors.

Fixes: #174

### DIFF
--- a/seedlet/functions.php
+++ b/seedlet/functions.php
@@ -150,39 +150,56 @@ if ( ! function_exists( 'seedlet_setup' ) ) :
 		);
 
 		// Editor color palette.
+		$colors_theme_mod = get_theme_mod( 'custom_colors_active' );
+		$primary          = ( ! empty( $colors_theme_mod ) &&
+								'default' === $colors_theme_mod ||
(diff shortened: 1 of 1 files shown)


allancole:
@jffng This should do the trick but the conditional code is pretty ugly, lol. If you could clean that up somehow, that’d be great :-)

allancole:
Also @kjellr, this can happen in a separate PR, but I really think we need to rename `--color-background-light` to `--color-tertiary` everywhere in the system for v1.0. Seedlet uses it as a lighter background color, but other child themes might want to use it differently so naming it something more vague will imply flexibility there. If we do it in v1.1, I fear it could be harder to unravel and require us to create extra fallbacks just for that one variable. Let me know what you think.

kjellr:
> If we do it in v1.1, I fear it could be harder to unravel and require us to create extra fallbacks just for that one variable. Let me know what you think.

Yeah, that seems doable. Would you mind filing an issue for it?

